PR #39, opened 2015-01-02
Userland fixes

slp:
These are a bunch of changes to make everything in Applications/util (except termcap and tget) build and run (at least, under z80pack).

A big part of this work has been removing duplicities between fuzix.h and other headers. For syscalls, I've checked the actual implementation in the Kernel itself to be sure about which arguments are the right ones.

It needs more testing and there still some improvements to be done, but I think this is on the right track.


EtchedPixels:
Pushed all the bits I merged and rebuilt all of userspace.

grep fails because it needs regexp, sed takes hours otherwise ok

PR #40, opened 2015-01-02
Noninteractive UCP

slp:
This implements a non-interactive mode for UCP, allowing us to use it inside scripts. I've also created a script for z80pack to automatically generate drivea.cpm (boot disk) and drivei.cpm (hard drive with utilities), using UCP this way.


EtchedPixels:
Much needed. Thanks.

I will tweak the use of /tmp/ to instead use local temporary files though, just so its safe.

PR #41, opened 2015-01-02
Filesys fix

slp:
Some small changes for fs_tab, getdev and newfstab which fix mounting non-root filesystems.


EtchedPixels:
Thanks, merged

PR #52, opened 2015-01-04
Introduce new "blkdev" layer for 512-byte sector partitioned drives.

willsowerbutts:
Hardware drivers are now responsible for the bare mimumum -- identify
the storage devices present, figure out how large they are, and provide
a function to read/write sectors.

The concept of "slices" goes away, instead we will use a separate
partition for each file system.

The blkdev layer handles PC-style MBR partition tables, both primary and
extended partitions, up to a maximum of 15 partitions per drive. The
partition numbering is chosen to be compatible with Linux.

Storage drivers no longer each require a separate entry in the device
switch table.

This includes patches for n8vem-mark4, p112 and msx2 all of which are using
devsd or devide.

### DIFF
--- a/Kernel/dev/blkdev.c
+++ b/Kernel/dev/blkdev.c
@@ -1,0 +1,149 @@
+#include <kernel.h>
+#include <kdata.h>
+#include <printf.h>
+#include <blkdev.h>
+#include <mbr.h>
+
+/*
+   Minor numbers
+
+   The kernel identifies our storage devices with an 8-bit minor number.
+
+   The top four bits of the minor identify the drive, allowing a maximum of
+   sixteen drives.
+
+   The bottom four bits of the minor identify the partition number. Partition
+   zero addresses "the whole drive", with no translation. Due to limitations
+   within the kernel only the first 32MB can currently be accessed through
+   partition zero.
+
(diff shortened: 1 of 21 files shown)


EtchedPixels:
Looks sane to me. Not all systems would use PC partition tables but the tie between the partition format and the PC one isn't tight. As an API it doesn't handle swap and raw which means I can't use it for example with IDE on the zx128 port as is, or for the trs80 disk controller. 

Not sure the extended handling works safely - it looks like it can loop.

So I think the features it needs would be
- pc versus other partition types (looks no problem)
- not looping on a bad partition
- automatically tagging a swap partition type so we don't swap on things evilly on sane boxes but don't need space to run a swapon() which we may not have
- swap/raw support - where it looks like if the API changes slightly it actually helps, as the single point can wrap the weirdnesses of the different call formats and just call the blkdev API for each block but pass in the type info as well.
- also looks like we get in a tangle if we want to support removable media ?

(I really would like to make raw support mandatory as it'll make execve work so much better)

None of these look like problems. I'm a bit concerned about code size but the really small boxes can still opt out of the feature and will be fine.

I'll merge this as its definitely the right direction.

PR #58, opened 2015-01-06
Use relative paths in FCC

slp:
Hi,

I think it would be nice having the ability of running FCC without installing it in "/opt/fcc". This changeset implements this option using relative paths and renaming 'tools' to 'bin' and 'libs' to 'lib'.


EtchedPixels:
libs is intentionally called libs because it will be several libraries in time and tools isn't just general binaries.

I agree it should be possible to decide where to put it (including within the tree) but that ought to be done by some kind of "make install" that puts the includes/libraries/tools where the user wants them to go including perhaps in a Build/ directory of the top directory ?

slp:
Sounds reasonable, but I'm not sure if installing FCC outside the working directory is something desirable. I, speaking for myself, prefer to keep everything self-contained.

Perhaps an alternative approach could be keeping the original names of both directories, while applying the changes on "fcc.c" for using relative paths, pointing to "../tools" and "../libs". That would allow to run FCC self-contained, until "make install" is implemented (and in that case, if it creates the same structure in "/opt/fcc", with binaries under "tools" and "crt0/c.lib" inside "libs", it would work anyway).

EtchedPixels:
Or how about making it support either environment variables or options for the tools path ? (lib is already covered I think with -L, include with -I)

Supporting something like FCC_INCLUDE_DIR FCC_LIBRARY_DIR, FCC_TOOLS_DIR would in general also be useful.

PR #60, opened 2015-01-08
devide: Flush write cache after each write (where supported)

willsowerbutts:


willsowerbutts:
OK, I will rewrite accordingly.

PR #62, opened 2015-01-09
Cache flush (take 2)

willsowerbutts:
How about something like this?

This patch requires platforms to change their device tables, only the two Z180 platforms are modified, so changes to other platforms would be required also.

Not completely happy with how _sync() and bufsync() call d_flush() but wanted to get your input at this stage in case this is not to your liking.

W

### DIFF
--- a/Kernel/dev/blkdev.c
+++ b/Kernel/dev/blkdev.c
@@ -120,6 +120,19 @@ int blkdev_write(uint8_t minor, uint8_t rawflag, uint8_t flag)
     return blkdev_transfer(minor, false, rawflag);
 }
 
+int blkdev_flush(uint8_t minor)
+{
+    blkdev_t *blk;
+
+    /* we trust that blkdev_open() has already verified that this minor number is valid */
+    blk = &blkdev_table[minor >> 4];
+
+    if(blk->flush)
+	return blk->flush(blk->drive_number);
+    else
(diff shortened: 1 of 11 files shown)


willsowerbutts:
Now using ioctl. Also has a tweak in devide to avoid sending a flush if we haven't sent any writes since the last flush.

EtchedPixels:
Not sure what has happened with this set of changes but they don't merge cleanly with the current tree (before or after I merged the later bugfix patch you posted)

willsowerbutts:
I will send you an updated series.

W

On Sun, Jan 11, 2015 at 12:49:16PM -0800, EtchedPixels wrote:
>Not sure what has happened with this set of changes but they don't merge cleanly with the current tree (before or after I merged the later bugfix patch you posted)
>
>
>---
>Reply to this email directly or view it on GitHub:
>https://github.com/EtchedPixels/FUZIX/pull/62#issuecomment-69510770

_________________________________________________________________________
William R Sowerbutts                                  will@sowerbutts.com
"Carpe post meridiem"                               http://sowerbutts.com
         main(){char*s=">#=0> ^#X@#@^7=",c=0,m;for(;c<15;c++)for
         (m=-1;m<7;putchar(m++/6&c%3/2?10:s[c]-31&1<<m?42:32));}

PR #66, opened 2015-01-11
Cache flush (take 3)

willsowerbutts:
Here we go again. Now based against alan/master, and tidied up into a series of commits that add the features progressively.


EtchedPixels:
Changed the ioctl setup a tiny bit and renamed it to match the Linux equivalent ioctl, and make it so root can use it from userspace.

PR #70, opened 2015-01-17
Exponential backoff at unsuccessful login attempts  and preliminary initscript support

erkinalp:
If five consecutive login attempts are unsuccessful, a backoff ranging from one minute to 512 minutes will be applied with this pull.

### DIFF
--- a/Applications/util/init.c
+++ b/Applications/util/init.c
@@ -25,7 +25,7 @@ void backoff(int);
 int main(int argc, char *argv[])
 {
     int fdtty1, sh_pid, pid;
-    char* rc_arg={"/etc/rc",NULL};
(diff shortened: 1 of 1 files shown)


erkinalp:
Issue #69 was based on previous revision, I have closed.

EtchedPixels:
Yeah I don't want to fiddle with the minimal init until we've got a proper init setup of some kind. its rather useful having something minimal right now.

erkinalp:
Zero environment in initscript shell spawning is intentional. We have not enough memory to hold two environment blocks simultaneously because Z80 and the likes are very RISC-ey compared to a DEC PDP, in turn, need more memory use for code.

PR #73, opened 2015-01-21
Simple script to build a root filesystem, plus tiny patch to ucp.

willsowerbutts:


willsowerbutts:
This automates the process that I previously executed by hand to build a root filesystem.

PR #77, opened 2015-01-23
Kernel: Remove the tty_writeready() function

willsowerbutts:
I added tty_writeready() to socz80-UZI as a way to get better
concurrency by causing a process writing to a busy UART port to sleep,
thus allowing a second process to run, and later waking up the sleeping
process once a UART interrupt arrives to indicate it is ready for
another byte.

This worked well enough on socz80-UZI with its 128MHz processor, but on
real Z80 hardware this scheme actually hurts performance at most
reasonable baud rates; the necessary overhead of sleeping the process,
switching to another process, taking an interrupt and switching back
takes much longer than just waiting for the UART to finish transmitting.

This patch removes tty_writeready in the core kernel and for each
platform. Note that an individual platform which wishes to retain the
old "sleep on busy UART" behaviour can still do so simply by
implementing this inside the platform tty_putc() routine.

### DIFF
--- a/Kernel/PORTING
+++ b/Kernel/PORTING
@@ -290,11 +290,6 @@ Copy the struct s_queue from another filled in for your devices/buffers.
 
 Provide putchar(c) if it is not already provided by your asm code
 
-bool tty_writeready(uint8_t minor)
-
-Returns true fi you can write bytes to this port (1..n). For things like
-video consoles just return true
-
 void tty_putc(uint8_t minor, char c)
 
 Write a character to the tty minor
--- a/Kernel/dev/z80pack/devtty.c
+++ b/Kernel/dev/z80pack/devtty.c
@@ -32,27 +32,17 @@ void kputchar(char c)
     tty_putc(1, c);
 }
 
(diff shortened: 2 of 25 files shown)


EtchedPixels:
I've pushed a tiny patch that does the need_resched() thing. See what you think

willsowerbutts:
OK, I agree this solution is sensible.

In the case of really slow devices we may end up spinning for quite a while before we psleep(). Maybe a simple tweak would be

#define need_resched() (nready != 1 && runticks)

ie end the time slice early?

EtchedPixels:
Umm

If its usually a win to spin and push bytes out then we should do that, if its worth a reschedule then presumably we should sleep. Is there really a half way house here ?

There is already another condition to check (flow contro) so we could make tty_waitready smarter and return 0/1/-1 'yes/no/sleep') and also factor in baud < threshold.

Short of trying a few ideas I've no idea which will work, but flow control one needs a fix - that I know about.

willsowerbutts:
I think having tty_writeready() return yes/spin/sleep is a great idea.

PR #84, opened 2015-01-24
Fix strerror, df

willsowerbutts:


willsowerbutts:
Snuck a tiny cosmetic fix to mbr in here too

PR #85, opened 2015-01-25
tty: sleep when reschedule required even if the drivers wants us to spin

willsowerbutts:


willsowerbutts:
Not a bug. Ignore me.

PR #86, opened 2015-01-25
Kernel: Bug fix for syscall_exec

willsowerbutts:
Convert block number to off_t before shifting, otherwise we're restricted to accessing only the first 64KB of the disk.


willsowerbutts:
I'm going to roll this up with some other changes in a new request.

PR #88, opened 2015-01-25
Fix count= option to dd, remove safety check from devio

willsowerbutts:


willsowerbutts:
PS would you like me to produce an updated patch that wraps this in #ifdef DEBUG?

EtchedPixels:
Yes I understand the problem and the trigger. I'm trying to work out the right cure. The other approach is to change readi for F_BDEV to do write the superblock back which is uglier.

I think the best change at this point is probably to only panic if its not the superblock. I don't think it can occur for any other block.

willsowerbutts:
I will look at this again.

EtchedPixels:
I have a think about it. If b_busy is turned from bool into a uint8_t then it can have both a busy flag and a pinned flag. Superblocks are pinned but not busy, normal disk I/O is busy but not pinned, and allocate/free checks both.

At least I think that is the proper fix for the problem I introduced putting the superblocks into the fs cache.

willsowerbutts:
Yes that is exactly what I was planning -- FREE, BUSY, SUPERBLOCK states.

W

On Mon, Jan 26, 2015 at 03:23:38AM -0800, EtchedPixels wrote:
>I have a think about it. If b_busy is turned from bool into a uint8_t then it can have both a busy flag and a pinned flag. Superblocks are pinned but not busy, normal disk I/O is busy but not pinned, and allocate/free checks both.
>
>At least I think that is the proper fix for the problem I introduced putting the superblocks into the fs cache.
>
>
>---
>Reply to this email directly or view it on GitHub:
>https://github.com/EtchedPixels/FUZIX/pull/88#issuecomment-71447078

_________________________________________________________________________
William R Sowerbutts                                  will@sowerbutts.com
"Carpe post meridiem"                               http://sowerbutts.com
         main(){char*s=">#=0> ^#X@#@^7=",c=0,m;for(;c<15;c++)for
         (m=-1;m<7;putchar(m++/6&c%3/2?10:s[c]-31&1<<m?42:32));}

PR #92, opened 2015-01-30
small fix devsd

geijoenr:
I have been observing a timeout when rx sectors that required a retry; this solves it.


geijoenr:
now this always skips a byte, I am not sure that is always good; please check will Will. S. if this is ok before merging.

EtchedPixels:
I see you closed this - is it waiting for something or wrong ? (confused)

geijoenr:
it works for me but I think is wrong, let me double check and will send again.

PR #93, opened 2015-01-31
n8vem-mark4: devsdspi - save a byte

willsowerbutts:
Based on your socz80 SD driver


willsowerbutts:
Hang on this has some crap in it. Let's spin it again.

EtchedPixels:
You can probably save two more by swapping the jr/call with a conditional call as well ;)

EtchedPixels:
Also on the sd subject see : https://github.com/EtchedPixels/FUZIX/pull/92

PR #96, opened 2015-01-31
msx2: devsd fix, rtc driver for rp5c01, international keyboard layouts

geijoenr:


geijoenr:
will do that. every byte counts.

PR #101, opened 2015-02-02
Kernel: tty_writeready: change return type from bool to new enum ttyready_t

willsowerbutts:


willsowerbutts:
Ugh. Broken. Hang about.

PR #102, opened 2015-02-02
Kernel: tty_writeready: change return type from bool to new enum ttyready_t

willsowerbutts:


willsowerbutts:
Rather unexpectedly this change saves 51 bytes. Unexpected bonus. On examining the assembler output I think it just tricks sdcc into generating better code in the monster tty_inproc() routine because that's the only thing that changes size.

PR #106, opened 2015-02-10
osh added

erkinalp:
Both ports of Thompson shell are added. sh6 and osh(version 20150115). Ideally we should choose one of them but for now this may remain as it is. Examples distributed with tarball not added.


erkinalp:
Unsuitable for inclusion. Closing.

PR #109, opened 2015-02-12
Kernel: readi() - when reading whole sectors, transfer them directly into user memory.

willsowerbutts:
Similar could be done for writei() if required.

### DIFF
--- a/Kernel/inode.c
+++ b/Kernel/inode.c
@@ -11,6 +11,8 @@ void readi(inoptr ino, uint8_t flag)
 	unsigned char *bp;
 	uint16_t dev;
 	bool ispipe;
+	off_t uostash;
+	usize_t ucstash;
 
 	dev = ino->c_dev;
 	ispipe = false;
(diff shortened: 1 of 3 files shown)


erkinalp:
On Z80, 6502 and 8086, there is no hardware privilege separation so user could already access system memory on these CPUs.

gardners:
While true that the basic versions of those CPUs have no privilege separation, there are some existing and new variants that do, including reimplementations in FPGA.  Thus is would be good to have the provision to support hardware privilege separation in the OS if it doesn't cause too much trouble.

EtchedPixels:
That reduces disk caching a lot so it's going to be a lose on bigger boxes and a win on small ones. I think its the right thing to do but with some kind of heuristic (eg 'total size of read > num cache blocks / 2') || O_DIRECT. That code path is also used for directory operations (see ch_link) and right now it happens not to go via your uncached route more by lack of optimization than anything else.

Also shows blk_op needs to become general I think.

gardners: the priv separation isn't a problem here. The request gets issued to the driver as a 'read to user space' so the driver is already handling it. The caller has also already validated the address range.

willsowerbutts:
Just pushed a patch for the u_sysio issue, and a second patch to make this feature a per-platform option, off by default but enabled on the Z180 platforms. Turning this on costs just under 200 bytes of code so it seems sensible for this to be optional.

Strongly agree the overloading of u_offset/u_count and the requirement to express block numbers as byte offsets are both unpleasant and cost code size. Would love something cleaner and more consistent. We could hoist blkdev up into the core Kernel.

I did consider the cache bypass issue, although I confess I had my 64K address space blinkers on at the time. The cache on this scale of machine will be tiny until we can push the data into other bank(s). Every time you cat a ~4KB+ file you end up essentially replacing the entire cache. I noted that executable code is also uncached.

I concluded it was probably a benefit to keep these transfers out of the cache and save the space for things being read in smaller units. I do think the right threshold for using direct transfers is 512 bytes -- userspace BUFSIZ is 512, a quick look in the utils confirms cat, cmp, dd, od and levee (at least) all use buffers of 512 bytes. If we made the threshold any higher we'd (virtually) never do it.

Side note, this was initially motivated by working on moving the CP/M emulator loading out of the kernel and into a userspace application. I felt it desirable to have the same ability to load the executable directly that execve() enjoys.

EtchedPixels:
I think the blkdev request structure at least belongs in kernel core for block type requests. It could actually be used for all requests but then it would have to live in the udata block.

Agreed on the bufsiz. Probably the logic is something like

if tiny box
     exclude (for now) - maybe once the blk stuff is tidier and its cheaper to do.
if small box
     include
if large box
     only do if O_DIRECT is set

For small boxes we will still get metadata cached (which is more valuable) and we'd get write then read caching which is the other usual case.

willsowerbutts:
Sorry about 0894aaf. It was just bringing this branch into sync. Now rebased against your current head. Are you happy to merge this branch? I think the blkdev merge should be done separately.

EtchedPixels:
Merged, and I'll merge a tweak to it to do O_DIRECT on bigger boxen.

I am now wondering if we have it backwards in the core code and the block layer requests should be made to look like the userspace ones given every driver has to deal with that anyway.

PR #111, opened 2015-02-12
Kernel: The platform rules.mk should define the CPU used

willsowerbutts:
This means that only the PLATFORM variable needs to be specified in
Kernel/Makefile, instead both PLATFORM and CPU in combination.


EtchedPixels:
Neat

PR #112, opened 2015-02-13
More powerful shell, zlib and zlib-based archiving utilities

erkinalp:
Sash(standalone shell) version 3.8-dir hardcoded to do ls, and modified dd slightly
Zlib
minizip

### DIFF
--- a/Applications/util/init.c
+++ b/Applications/util/init.c
@@ -121,7 +121,7 @@ int login(char *ttyname)
 
             /* loop until a valid user name is entered
              * and a shell is spawned */
-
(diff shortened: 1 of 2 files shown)


EtchedPixels:
This seems to be have tons of stuff in it, and none of it actually ends up with a buildable sash or zlib for Fuzix.

The only thing I've taken from it is the special casing of pid 1 kill, which I've fixed properly in the kernel instead.

erkinalp:
Do you want any pieces of sash to stripped? There are too many builtins.

PR #114, opened 2015-02-13
Better root device name parsing

willsowerbutts:
With these changes you can specify (eg) "hdb1" or "fd0" at the boot device prompt (or on the command line). Backward compatible with the old numeric parsing.

### DIFF
--- a/Kernel/Makefile
+++ b/Kernel/Makefile
@@ -17,7 +17,7 @@ ROOT_DIR := $(shell dirname $(realpath $(lastword $(MAKEFILE_LIST))))
 
 all:	fuzix.bin
 
-# Platform propertyies
+# Platform properties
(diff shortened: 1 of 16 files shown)


EtchedPixels:
Merged although I will tidy up the ifdefs a bit.

As several platforms use the same bootdevice rules for fd/hd maybe that ought to be a standard helper lib too ?

willsowerbutts:
Yes. 

How about, in config.h, delimited list of device names in order, a # indicates a letter suffix.

#define BOOTDEV_NAMES "hd#|fd"
#define BOOTDEV_NAMES "rd|fd|hd#"

and a simple matching loop that counts up device numbers. The  Platform provides the string in config.h, the new version could go into Kernel/start.c

willsowerbutts:
I will write this up later.

PR #115, opened 2015-02-14
Kernel: Generic boot device parsing method

willsowerbutts:
I think this should work for any platform that uses sensible device naming.


willsowerbutts:
Did a few forced pushes to incorporate code size improvements; 33c388e is the final version.

PR #118, opened 2015-02-15
Ask old password when root is changing its own password

erkinalp:


willsowerbutts:
How is root supposed to change their password when they forget it?

erkinalp:
By deleting its password from /etc/passwd directly.

willsowerbutts:
So, if root can just change /etc/passwd directly, what does this change achieve?

erkinalp:
Discouraging root to forget its password. Anyway, these small boxes are pseudo-multiuser at most :)

willsowerbutts:
It's more of a roadblock than discouragement :) I would prefer to stick with the tradition (root is never prompted for the current password)

By the way, thank you for your work on porting a better shell. Something with globbing, pipelines and redirection would be great.

EtchedPixels:
The existing behaviour is traditional so should IMHO be kept

PR #123, opened 2015-02-17
Single user mode added

erkinalp:
Differences with muiti-user build include:
* Always root
* No password asked at login
* Dummy setuid,setgid

Single-user build may be useful for limited ROM situations as some utilities will not be needed such as `passwd`, `adduser`, `deluser`, `chown`, `chgrp`.

### DIFF
--- a/Applications/util/init.c
+++ b/Applications/util/init.c
@@ -121,7 +121,7 @@ int login(char *ttyname)
 
             /* loop until a valid user name is entered
              * and a shell is spawned */
-
+#ifndef SINGLE_USER
             for (;;) {
                 putstr("login: ");
                 while (read(0, buf, 20) < 0);    /* EINTR might happens because of the alarm() call below */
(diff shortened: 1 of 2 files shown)


EtchedPixels:
I don't understand why you need any kernel changes for this. You start off as root/root so they can simply be omitted.

The init bit maybe makes sense but could also be automated. Eg if before the putstr("login:") you did

pwd = getpwnam(root);
if (!pwd->pw-passwd[0])
     spawn(pwd);
else {
  putstr("login: ")

etc...

I'd rather avoid ifdefs

erkinalp:
Its purpose is to have an _inherently single-user_ build as opposed to an optional single-user boot mode. This is the reason for kernel patch.

EtchedPixels:
Why do I want one ? It serves no useful purpose

erkinalp:
Some space can be saved with utilities left out in a single-user build.

EtchedPixels:
You can leave them out anyway..

erkinalp:
Closing upon maintainer's request. Forks may make use of it.

PR #139, opened 2015-03-13
Updated build system WIP

davidgiven:
Hello,

I was trying to beat my MSP430 port into shape the other day and got frustrated at the build system, so thought I'd try to overhaul it; I may have gotten a little carried away...

This is a WIP of a replacement build system. The only kernel I've done so far is the msx2, and I haven't been able to verify that it works yet, but all the rest builds. Features include:

- builds everything; kernel, applications, libc, file system image, host tools
- proper dependency tracking of everything, including headers
- builds everything in a single pass (which means no recursive makefiles, and much faster incremental builds)
- fully parallelisable
- out-of-tree object files; different platforms' objects live in different subdirectories, so you can have multiple platforms built at the same time without their object files stepping on each other
- modular; all the target-specific compiler settings are isolated in their own file --- which means all the SDCC horror is abstracted away. Adding a new toolchain should be really easy
- preprocessable assembler files

Plus, some of the logic that's been dealt with in helper scripts is now done in the makefiles (because then the makefiles can reason about them). e.g. fcc is no longer necessary.

sdcc-rules.mk is unnecessarily complex, because SDCC. A gcc-based architecture (like my msp430) should be way simpler.

There's some documentation in README.build.mk.

Is any of this of use to you? i.e. would you like me to keep working on this? (If so I would suggest *not* accepting the pull request as I don't know whether I've broken the old build system (because I never made it work). I'd appreciate guidance; I'm not good with git.)Hello,

I was trying to beat my MSP430 port into shape the other day and got frustrated at the build system, so thought I'd try to overhaul it; I may have gotten a little carried away...

This is a WIP of a replacement build system. The only kernel I've done so far is the msx2, and I haven't been able to verify that it works yet, but all the rest builds. Features include:

- builds everything; kernel, applications, libc, file system image, host tools
- proper dependency tracking of everything, including headers
- builds everything in a single pass (which means no recursive makefiles, and much faster incremental builds)
- fully parallelisable
- out-of-tree object files; different platforms' objects live in different subdirectories, so you can have multiple platforms built at the same time without their object files stepping on each other
- modular; all the target-specific compiler settings are isolated in their own file --- which means all the SDCC horror is abstracted away. Adding a new toolchain should be really easy
- preprocessable assembler files

Plus, some of the logic that's been dealt with in helper scripts is now done in the makefiles (because then the makefiles can reason about them). e.g. fcc is no longer necessary.

sdcc-rules.mk is unnecessarily complex, because SDCC. A gcc-based architecture (like my msp430) should be way simpler.

There's some documentation in README.build.mk.

Is any of this of use to you? i.e. would you like me to keep working on this? (If so I would suggest *not* accepting the pull request as I don't know whether I've broken the old build system (because I never made it work). I'd appreciate guidance; I'm not good with git.)


Applications/build.mk
```diff
@@ -0,0 +1,104 @@
...
+$(OBJ)/Applications/util/date:     $(OBJ)/Applications/util/date.$O
+$(OBJ)/Applications/util/dd:       $(OBJ)/Applications/util/dd.$O
+$(OBJ)/Applications/util/decomp16: $(OBJ)/Applications/util/decomp16.$O
+$(OBJ)/Applications/util/df:       $(OBJ)/Applications/util/df.$O
+$(OBJ)/Applications/util/dirname:  $(OBJ)/Applications/util/dirname.$O
+$(OBJ)/Applications/util/dosread:  $(OBJ)/Applications/util/dosread.$O
+$(OBJ)/Applications/util/du:       $(OBJ)/Applications/util/du.$O
+$(OBJ)/Applications/util/echo:     $(OBJ)/Applications/util/echo.$O
+$(OBJ)/Applications/util/ed:       $(OBJ)/Applications/util/ed.$O
+$(OBJ)/Applications/util/false:    $(OBJ)/Applications/util/false.$O
```
erkinalp: I have added factor. Can you add it into new Makefile?

EtchedPixels:
A build system would be a wonderful thing - the current setup is very much nailed together

Applications/build.mk
davidgiven: Done (although I had to fix some bugs in it; did it ever work?).

erkinalp: Should work with #141 merged.

erkinalp:
David, I see traces of mercurial in your commit messages, can you point me to mercurial repository?

davidgiven:
@erkinalp: I'm actually using hg-git. I've never got my head around the way git works and this seems to interoperate pretty well. (Although branches are a little awkward: I have to create them on github and pull to hg where they show up as bookmarks. Creating bookmarks on hg and trying to push them just gets thing (including me) confused.)

Also, I didn't realise that .hgtags was being pushed, dammit.

Try:

hg clone git+ssh://git@github.com/davidgiven/FUZIX.git 

(Replace path with that of your own repo; note the syntax is different from git's.)

davidgiven:
Incidentally, this change is getting stupidly big so I'm going to try and split it up into smaller pull requests.

erkinalp:
@davidgiven: Mercurial has a different change tracking than git's. Git has a content-based tracking, mercurial has a file-oriented repository-based one. Therefore Mercurial branches will not match git's. Bookmarks corresponds to tags in git.

davidgiven:
This is being reworked to actually, you know, work.

PR #145, opened 2015-03-15
Libc tweaks

davidgiven:
This is a whole bunch of relatively minor libc changes:

- prototyping and function signature changes to be more consistent and standard
- replace the unused Z80 versions of the string functions with portable C versions (stolen from the ACK; 3-clause BSD licensed); Z80 platform use the copies from the compiler library anyway, and having C versions is required by platforms that don't have these in their compiler library
- some relatively minor changes to printf and scanf to reduce code size (by a few hundred bytes).

### DIFF
--- a/Library/include/stdlib.h
+++ b/Library/include/stdlib.h
@@ -67,7 +67,7 @@ extern int atexit __P((atexit_t));
 
 extern char *crypt __P((char *__key, char *__salt));
 
-typedef int (*cmp_func_t) __P((void *, void *));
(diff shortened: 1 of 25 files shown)


davidgiven:
*cough*

Pay no attention to the commit behind the curtain!

EtchedPixels:
Merged - please confirm MIT v BSD header on strlen.c ?

davidgiven:
Dammit, it should be BSD. (Copy-and-pasted from the wrong file.) Let me fix that.

PR #148, opened 2015-03-16
Smaller ctype implementation

davidgiven:
This replaces the existing table-based ctype system with a slower but smaller one based on doing comparisons in code. This reduces the size of some binaries, mostly due to not having the big 256-byte lookup table, but also due to commoning out code.

For 6502, the biggest winners are wc, ed and cut, all of whom get 250-350 bytes smaller; typically it's about 100 bytes. And of course programs that don't call ctype don't save anything. It's probably possible to shave a few more bytes off by changing the prototypes to 'char isdigit(char c)', but I'm pretty sure that violates the C standard.

Note that this is largely untested --- things build, but I don't have any way to run anything yet, I haven't been able to verify that it all still works. Plus it's late. Probably best to check my logic...


erkinalp:
> First obvious problem is that the ranges for ctype according to the spec are defined as
> EOF and 0-255, so you can't cast them to a char. The existing ones are also I think wrong in this > area in some spots. Then again if 255 and EOF have the same value for each test I guess casting to > uchar is safe.
Short is definitely safe, but requires a cast to int(which will matter on future 32-bit builds).

EtchedPixels:
I had a few minutes over lunch so I've taught fcc to define something on Z80 so we can tell high optimization levels and pick a ctype scheme to use

davidgiven:
Re unsigned: yup, fixing. (Actually makes the code *smaller*. tail shrinks from 14730 bytes to 14690 on Z80, 15712 to 15702 on 6502; other functions change similarly, and AFAICT nothing gets bigger.)

Re lookup tables: that's neat! But then I implemented it, and the bad news is that while tail gets smaller (14690 to 14658 on Z80), most of the other apps actually get bigger --- on the 6502 it goes up to 15727, which is worse than the original version. I wasn't expecting that at all. I put the lookup table in its own source file, which means that it'll be used by the various functions and won't be duplicated.

The new xdigit.o, which is the most complex function, is 29 bytes of Z80 code while the previous one is 52. That's only 23 bytes smaller, which means the saving is eaten up by the 32-byte lookup table. That table is shared by the other functions so it *should* be smaller in practice but it seems not to be. Not quite sure why.

Eventually I realised I could inline the simpler functions in ctype.h, because that saves having to link the function, having to generate the calls every time, having to convert the value into an int (because most people will be using these in conditionals anyway), *and* as the value's already live it saves there, too. tail gets down to 13860 bytes on Z80 (15691 on 6502), which is nearly a kilobyte smaller than the original! Looking at the code this seems to be because in the original code sdcc is doing a 16-bit pointer add when dereferencing __ctype, even when it doesn't need to:

   0085 7E            [ 7]  133         ld      a,(hl)
   0086 DD 77 F2      [19]  134         ld      -14 (ix), a
   0089 DD 77 F7      [19]  135         ld      -9 (ix),a
   008C C6r00         [ 7]  136         add     a,#<(___ctype)
   008E DD 77 F8      [19]  137         ld      -8 (ix),a
   0091 3Es00         [ 7]  138         ld      a,#>(___ctype)
   0093 CE 00         [ 7]  139         adc     a, #0x00
   0095 DD 77 F9      [19]  140         ld      -7 (ix),a
   0098 DD 6E F8      [19]  141         ld      l,-8 (ix)
   009B DD 66 F9      [19]  142         ld      h,-7 (ix)
   009E CB 5E         [12]  143         bit     3,(hl)
   ...etc..

vs

                                                  ld h, (hl)
   006F 7C            [ 4]  131         ld      a,h
   0070 EE 80         [ 7]  132         xor     a, #0x80
   0072 D6 B0         [ 7]  133         sub     a, #0xB0
   0074 38 0B         [12]  134         jr      C,00164$
   0076 3E 39         [ 7]  135         ld      a,#0x39
   0078 94            [ 4]  136         sub     a, h
   ...etc...

Again, logic needs checking, etc.

davidgiven:
*facepalm*

Yes, you're absolutely right. I should have spotted that. That's not going to work. Shame, as it was a huge size improvement.

Also, you know what this needs? A test suite.

My original change (with the fixes) is still an improvement over the table-based for SDCC ad 6502 (and also, surprisingly, for the MSP430, a modern register-centric 16-bit architecture I've been doing test builds on), but I'll resubmit that.with some cleanup. Feel free to reject this PR.

EtchedPixels:
I'll reject it once we have a final solution - right now its a rather good log of the discussion.

For 6502 I think I may just give in and rewrite isfoo in asm if I can beat the compiler.

Your defines look right for z80 if you turn them into inline and they come out nice and tight (SDCC inlining isn't actually that badly done).

davidgiven:
Well, I wrote a unit test for the ctype stuff.

That patch I sent the other day? So many bugs. Sorry, that was really embarrassing. The unit test should fix that.

So, SDCC inline: I converted to use static inline (patch included), but on compilation, the file sizes were a total disaster --- every file was about 2kB bigger.

Investigation shows that this is because while sdcc does do a good job of inlining, it also includes function bodies for every inline function, even though they're not called. (ref: https://sourceforge.net/p/sdcc/feature-requests/414/)

Suggestions? So far the best is f8e9323 above, which is simply out-of-lining the code. Although I'll need to backport the unit test and fix the bugs. Unless you know a way to make sdcc behave. I really want that extra kilobyte of space...

EtchedPixels:
Oh well.. out of line it is then

davidgiven:
Okay, done. It's not as small as I'd like, but it's still a decent size improvement. The figures for tail are now 14682 bytes for Z80, 15679 bytes for 6502, down from 14849 and 15866.

I managed to clean up the history a bit; that last commit is a merge from master and is a complete patch for the changes. Let me know if you'd like me to produce a clean one-commit PR.

erkinalp:
All these comments will be gone if you force-push over.

EtchedPixels:
Sweet

davidgiven:
Yay. Thanks!

PR #166, opened 2015-04-24
PPIDE for Zeta 2 SBC

willsowerbutts:
A number of changes. Most interesting one is PPIDE support for the Zeta 2 SBC.


willsowerbutts:
Ah that would be better, yes. Will send patch.

PR #199, opened 2015-06-04
Some more cc65 workarounds

davidgiven:
cc65 makes me sad (but the apps all build now, including sh and levee). Also includes a fix for a mismatched function prototype that was causing gcc to blow up.


EtchedPixels:
You could always fix cc65 ;-)

PR #203, opened 2015-06-08
gets_s fix

erkinalp:
gets_s() should act as no-op on *str if exceeded buffer while input according to ISO C.


EtchedPixels:
Wrong in several ways. I'll do the trivial corrections and C11 compliance properly at some point

erkinalp:
What sense is it wrong in? Stopping as soon as `n + 1` characters entered?

EtchedPixels:
Where do we start
- allocating a ton of memory on the stack
- doing pointless memory copies
- not setting erange
- not correctly handling overlong lines (data must be consumed)

So I made the simple trivial changes instead

erkinalp:
> doing pointless memory copies
To get a well-defined behavior without messing with full constraint-handler setup.

PR #204, opened 2015-06-09
readline added

erkinalp:
Readline library from Yorkie Neil added. Might be useful for shell.

### DIFF
--- a/Library/readline/readline.c
+++ b/Library/readline/readline.c
@@ -1,0 +1,93 @@
+
+//
+// readline.c
+//
+// Copyright (c) 2014 Yorkie Neil <yorkiefixer@gmail.com>
+//
+
+#include <string.h>
+#include <stdlib.h>
+#include <stdio.h>
+#include "readline.h"
+
+readline_t *
+readline_new(char * buffer) {
+  readline_t * rl = (readline_t*) malloc(sizeof(readline_t));
+  size_t len = strlen(buffer);
+
+  rl->cursor = 0;
+  rl->line = 0;
+  rl->buffer = (char*) malloc(len);
+  memset(rl->buffer, 0, len);
+  memcpy(rl->buffer, buffer, len);
+  return rl;
+}
+
+char *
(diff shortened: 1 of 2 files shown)


Library/readline/readline.c
```diff
@@ -0,0 +1,93 @@
...
+//
+
+#include <string.h>
+#include <stdlib.h>
+#include <stdio.h>
+#include "readline.h"
+
+readline_t *
+readline_new(char * buffer) {
+  readline_t * rl = (readline_t*) malloc(sizeof(readline_t));
```
Zardoz89: Perhaps it's me, but I have the impression that this library takes too assumptions about that malloc/calloc can't fail.

Library/readline/readline.c
```diff
@@ -0,0 +1,93 @@
...
+  size_t len = strlen(buffer);
+
+  rl->cursor = 0;
+  rl->line = 0;
+  rl->buffer = (char*) malloc(len);
+  memset(rl->buffer, 0, len);
```
Zardoz89: Why memset when you are going to overwrite with a memcpy of the same size ?

erkinalp: It is a dependency-break idiom in superscalar pipelined arches, unnecessary in our case. I will strip it in a later version.

Library/readline/readline.c
erkinalp: *NULL is reset vector on Z80 family so cannot harm any code or data that way on Z80.

EtchedPixels: Not always true, and since you are doing multiple bytes you then scribble over the vectors.

EtchedPixels:
We have no use for this code in anything in tree or is it part of SVID

erkinalp:
Readline is a dependency for many REPLs despite not being part of SVID or SUS.

Zardoz89:
I'm testing it against GCC .  I find that gives a warning about a bad inlined function, plus one function is broken. I'm trying to fix it...

PR #205, opened 2015-06-09
Readline added

erkinalp:
#204 plus fix for nonovercommit malloc().

### DIFF
--- a/Library/readline/readline.c
+++ b/Library/readline/readline.c
@@ -1,0 +1,93 @@
+
+//
+// readline.c
+//
+// Copyright (c) 2014 Yorkie Neil <yorkiefixer@gmail.com>
+//
+
+#include <string.h>
+#include <stdlib.h>
+#include <stdio.h>
+#include "readline.h"
+
+readline_t *
+readline_new(char * buffer) {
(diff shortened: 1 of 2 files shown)


Zardoz89:
I have curiosity. From were comes these idiom to do memset/memcpy to do dependency-break ? I can't find anything about it. 

Also, check this : https://github.com/Zardoz89/readline

erkinalp:
@Zardoz89: It is not a C idiom, it is a hardware-level idiom. Some CPU architectures like AMD64 have superscalar schedulers. You break the dependency into previous access of that memory location if you zero it. The CPU scheduler continues to dispatch your operations in respect to previous values if you do not zero them. It is a pessimization on non-pipelined non-superscalar architectures.

Zardoz89:
Also, you know that they are self-depracting the library ? 
https://github.com/clibs/readline/pull/2#issuecomment-110528928
https://github.com/clibs/clib/issues/21#issuecomment-33294863

EtchedPixels:
Most modern processors are way smarter than to need the memset() trick, also good compilers and C libraries should deal with it for you.

PR #207, opened 2015-06-10
Use precompiled sdcc for Travis builds

davidgiven:
Rather than downloading and building sdcc ourselves, use the precompiled version in my PPA. This reduces build time from about 8 minutes to about 5.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,13 +4,7 @@ cache:
 before_install:
     - sudo add-apt-repository ppa:david.given/ppa -y
     - sudo apt-get update -q
-    - sudo apt-get install cc65
-    - mkdir -p dependencies
-    - cd dependencies
-    - sudo apt-get --no-install-recommends install libboost-all-dev
-    - git clone --depth 1 --branch patched https://github.com/davidgiven/sdcc.git
-    - (cd sdcc/sdcc && ./configure --prefix=/usr/local --disable-pic14-port --disable-pic16-port)
-    - (cd sdcc/sdcc && make -j4 > /dev/null && sudo make install > /dev/null)
+    - sudo apt-get install cc65 sdcc
     - export PATH=$PATH:/usr/local/bin
     - cd ..
 language: c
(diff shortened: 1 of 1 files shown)


erkinalp:
Travis CI seems inactive on EtchedPixels.

davidgiven:
Yes, it's not set up there yet (it's not my repository so I can't set it up). But it's trivial to set up for your own, which means it'll do automatic test builds before sending off pull requests. Just log in to travis-ci, refresh your repositories, and switch your Fuzix project to 'on'.

erkinalp:
Do I have to pay for it?

Zardoz89:
It's free. But the owner (user or team member) of the github repository is who can activate it agains the github repo. In other words @EtchedPixels  .
We use it on Trillek and it's really a great tool to see if something gets broken on a possible merge.

EtchedPixels:
I'm waiting for it all to settle down somewhat first, and for some of the problem cases to be dealt with (eg you need sdcc 3.4.2 or higher to build a valid kernel for some platforms, and I don't think 3.4.0 builds all the user space validly either).

I'm hoping the pending 3.5.0 has everything we need in it and works nicely.

Zardoz89:
At least these travis file give us a basic recipe how to build fuzix

erkinalp:
We still have minutes to half an hour build files. We should have better seconds of incremental builds before continuous integration.

EtchedPixels:
sdcc is very slow so that isn't going to happen

erkinalp:
Then transistion into a faster compiler. Do we have a target where SDCC is the only existing C compiler?

EtchedPixels:
There is no remotely comparable Z80/Z180 compiler.

erkinalp:
> remotely comparable

What aspect? Speed? Optimisation?

Zardoz89:
z88dk ?

EtchedPixels:
botth - and no z88dk isn't even in the running. Plus z88dk is slowly moving towards using sdcc.

PR #208, opened 2015-06-13
Add MSP430 Travis support

davidgiven:
This sets up the compiler and autobuilds the MSP430 userland. There's no kernel for it yet, so its main job is to check to make sure everything builds on a reasonably standard gcc.


davidgiven:
Oops, misread the build status --- this doesn't actually work.

PR #212, opened 2015-06-17
Tac

erkinalp:
Backwards version of cat.


erkinalp:
Still can I replace them by UNIX library calls and will do so in a suitable time.

erkinalp:
I will also add a utility which does the job of `tac | rev` in a single invocation.

PR #217, opened 2015-06-27
zeta-v2: Make _trap_reboot compatible with UNA BIOS

willsowerbutts:


willsowerbutts:
John has come up with a much better fix that remaps ROM to all of the pages just like a "real" hardware reset. I'll write up a better patch for that today.

PR #220, opened 2015-07-11
dd: a fix and a feature

erkinalp:
Now initializing locals statically at declarator, therefore saving initialization code. Also strfry=NNN option added which displaces data within given block.

### DIFF
--- a/Applications/util/dd.c
+++ b/Applications/util/dd.c
@@ -76,7 +76,7 @@ void main(int argc, char *argv[])
     long outtotal;
     long inmax=0;
     char *buf;
-    size_t mixupblk=1;
+    size_t mixupblk=0;
 
     while (--argc > 0) {
 	str = *++argv;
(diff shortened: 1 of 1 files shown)


EtchedPixels:
Saves no code at all. The compiler will generate the same code for both for auto variables.

And we don't need silly options on a small machine.

erkinalp:
Will staticify the initialized variables and re-make the PR. OtOH, strfry=nnn is a useful secure removal tool in the situations if you are allowed to owerwrite a file with garbage data but not securely erase it. Shuffle as a proxy for erase.

EtchedPixels:
overwriting a file doesn't erase it on a lot of media types and you could just as well overwrite it using yes and dd, plus anyone using Fuzix on an 8bit micro for secure activities is crazy,

erkinalp:
> doesn't erase it on a lot of media types

It does in those Unix is originally intended for, i.e. Floppy, tape, HDD, puncher, line printer.

> using Fuzix on an 8bit micro for secure activities is crazy

FUZIX does have 16/32 bit targets too, those are usable in embedded applications. Not all FUZIX targets are vintage 8-bit resurrecters.

EtchedPixels:
Floppy needs a bulk eraser or the right patterns and  lot of care. Tapes need to be bulk erased or via an erase head. Hard disks do block reassignment so cannot safely be erased this way. Modern ones have special commands for it. Punched paper tape is erased by putting it in a shredder and line printer output likewise.

So it's not actually any use.

PR #230, opened 2015-08-24
Fforth

davidgiven:
So this was going to be a quick hack, but it got a bit carried away. It's not *quite* done, but I'm off on holiday for a bit so I thought I'd get this to you now.

This is a lightweight ANSish Forth interpreter. It produces about a 10kB binary on the MSP430 --- I haven't tried the other platforms yet. It's indirect threaded and should be pretty portable. It does need C99. There are crude and untested system call bindings to a lot of the Fuzix system calls.

It's almost completely untested and I haven't run it on anything other than my PC yet, but it passes most of an ANS Forth conformance tests that I found.

It's a single C file, and the implementation is entertainingly evil.


davidgiven:
...also I now notice that my tree manipulations to try and get this out of my MSP430 directory seem to have done some really odd things, so you probably don't want to merge this.

davidgiven:
Okay, fixed --- the history should make sense now.

PR #231, opened 2015-08-26
Cmdline

beretta42:
Proposed kernel argc and argv.  Not ready for prime-time.  Kernel modules then could scan kargv for boot params.  There should probably be some atoi conversion routine in here as well?

### DIFF
--- a/Kernel/platform-coco3/Makefile
+++ b/Kernel/platform-coco3/Makefile
@@ -1,5 +1,5 @@
 
-CSRCS = ttydw.c mbr.c
+CSRCS = ttydw.c mbr.c cmdline.c
 CSRCS += devices.c main.c libc.c devsdc.c
 
 DSRCS = ../dev/devdw.c ../dev/blkdev.c ../dev/devide.c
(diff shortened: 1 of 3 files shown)


EtchedPixels:
Other comment. It's probably better that instead of an array you just parse it in a discard file for the platform as you find each token. That way everything is thrown away easily.

EtchedPixels:
Something like

 do {
   while(*p && *p++==' ');
   s = p;
   while(*p && *p++!=' ');
   if (*p) *p++=0;
   if (*s && !platform_option(s))
      add_to_init(s);
} while(*p);

perhaps ?

beretta42:
sweet.


On Wed, Aug 26, 2015 at 6:16 PM, EtchedPixels <notifications@github.com>
wrote:

> Something like
>
> do {
> while(
>
> *p && *p++==' '); s = p; while(*p &&
> *p++!=' '); if (*p)
> *p++=0; if (*s && !platform_option(s))
> add_to_init(s);
> } while(*p);
>
> perhaps ?
>
> —
> <https://github.com/EtchedPixels/FUZIX/pull/231#issuecomment-135193708>.
>



-- 
Brett M. Gordon,
beretta42@gmail.com

PR #234, opened 2015-09-17
coco3: boot: commandline works with REM token.

beretta42:
Old code only worked with the ' comment in BASIC. Now the REM token will also work for passing command line to kernel. (Thanks William Astle)


erkinalp:
Passing kernel parameter through BASIC interpreter?

beretta42:
erkinalp:  Yes.   Technically, the code change was to the stand alone ASM booter... it loads up the kernel image, and grabs the "command line" from BASIC and copies it to where FUZIX expects it.
See Kernel/platform-coco3/README.  Cheers!

PR #235, opened 2015-09-17
CoCo3: check machine for 512k ram.

beretta42:
panic if machine only has 128k.


beretta42:
sorry non ready for prime time..

beretta42:
Something smells with this code.  Calling panic() before the system is initialized feels dirty.

EtchedPixels:
Possibly the check needs to happen really early before you blow away the basic/rom etc so you can print it and fall back to basic ?

Its a general problem - if you find something fails early in startup its hard to report - not just a Fuzix problem (Linux sucks at it too 8))

PR #239, opened 2015-09-23
blkdev: increment data address and lba each time through read/write loop

beretta42:
The main sector transfer loop, was just sending the low-level driver the same address and lba every loop....

### DIFF
--- a/Kernel/dev/blkdev.c
+++ b/Kernel/dev/blkdev.c
@@ -126,6 +126,8 @@ static int blkdev_transfer(uint8_t minor, uint8_t rawflag)
             goto xferfail;
         blk_op.nblock -= n;
         count += n;
+	blk_op.addr += n * BLKSIZE;
+	blk_op.lba += n;
     }
 
     return count; /* 10/10, would transfer sectors again */


atsidaev:
sdcc cannot compile this instruction. Since **blk_op.addr** has void* type, the following error occures on all z80 ports:
```
../dev/blkdev.c:129: error 66: attempt to allocate variable of unknown size 'iTemp130'
```
Checked both with patched 3.4.2 and trunk sdcc versions.

beretta42:
Bah!  missed that one.  I forget GCC allows for void pointer arithmatic.
Sorry.  Let me cast the void* to a char*, I'll submit a new pull request.

On Sat, Sep 26, 2015 at 8:49 AM, Alexander Tsidaev <notifications@github.com
> wrote:

> sdcc cannot compile this instruction. Since *blk_op.addr* has void* type,
> the following error occures on all z80 ports:
>
> ../dev/blkdev.c:129: error 66: attempt to allocate variable of unknown size 'iTemp130'
>
> Checked both with patched 3.4.2 and trunk sdcc versions.
>
> —
> <https://github.com/EtchedPixels/FUZIX/pull/239#issuecomment-143433751>.
>



-- 
Brett M. Gordon,
beretta42@gmail.com

EtchedPixels:
blk_op.addr should probably be uint8_t * rather than casting it.

PR #241, opened 2015-09-25
coco3: bugfix: mbr.c adjust offsets/lens for 512b blocks.

beretta42:
Needed to convert offsets / lens in the partition table to 512b block addressing, as our partition table can be formated in 256, 512, 1024.... etc...


EtchedPixels:
Is the COCO3 partition format documented anywhere, and is it used on older COCO boxen - just asking as at some point I should merge the COCO 64K port and right now it doesn't do partitions but fixed disk slices.

PR #243, opened 2015-09-25
coco3: bugfix: don't set PAL mode for the coco3.

beretta42:
This was glitching out my TV.  I forgot I live in N. America. :)

### DIFF
--- a/Kernel/platform-coco3/coco3.s
+++ b/Kernel/platform-coco3/coco3.s
@@ -138,7 +138,7 @@ b@	sta	,x+
         ;; set temporary screen up
 	ldb	#%01000100	; coco3 mode
 	stb	$ff90
-	ldb	#%00001100	; text / 8 lines per char row
+	ldb	#%00000100	; text / 8 lines per char row
 	stb	$ff98
 	ldb	#%00010100	; 80 column mode
 	stb	$ff99


erkinalp:
I will add an #ifdef and send a PR for PAL/NTSC thing.

beretta42:
Yeah, that's probably a better strategy.  I'm unfamiliar with the CoCo's ability to handle PAL.  This bit seems to do nothing to the RGB port, but It does affect the Composite Out.

EtchedPixels:
Does the basic firmware not leave this information somewhere (ROM difference etc ?) so it can autodetect. Failing that can it not be autodetected by something else - eg COCO1/2 invert the hsync
direction to the PIA, and in all cases the cpu clocks per hsync will differ.

beretta42:
I don't think this setting is of any use, except when using a PAL CoCo3 via the RF modulator.  According to these sources, it looks like only users of PAL CoCo3 RF output would benefit.  Is it worth the complexity of checking ROM?  It does nothing to the RGB signal, but messes up NTSC coco3's composite out.

http://www.cocopedia.com/wiki/index.php/PAL_Coco_Differences
http://tandycoco.com/coco3.shtml

erkinalp:
Do we have a chance to pass this PAL/NTSC setting on kernel so it can autoset line count. In NTSC, 21(15 if CJK) lines are readable, however, in PAL, 28(20 if CJK) lines are readable according to my observation on a decent PAL TV with teletext font.

beretta42:
The CoCo3 only produces 192, 200 and 225 lines progressive.  So no reason
there, but maybe we should pass this to the kernel to help a VSYNC based
clock, though...

On Sat, Sep 26, 2015 at 6:36 AM, Erkin Alp Güney <notifications@github.com>
wrote:

> Do we have a chance to pass this PAL/NTSC setting on kernel so it can
> autoset line count. In NTSC, 21(15 if CJK) lines are readable, however, in
> PAL, 28(20 if CJK) lines are readable according to my observation on a
> decent PAL TV with teletext font.
>
> —
> <https://github.com/EtchedPixels/FUZIX/pull/243#issuecomment-143418286>.
>



-- 
Brett M. Gordon,
beretta42@gmail.com

erkinalp:
@beretta42: I was mentioning of text lines, not scanlines yet this is helpful for font selection. In PAL, this amounts to 10x8 per character; in NTSC; this amounts to 10x9 per character using given grid.

EtchedPixels:
Just about all of the planet except North America is PAL not NTSC so it's probably best autodetected somehow or left how basic left it.

@erkinalp it makes no difference to the display the COCO generates extra blank space on PAL displays not more lines (in part I guess because they had to glue the 6847 to a PAL display so the only way to do it was to use extra blanking counters (see the COCO technical reference manual). It also means the games don't break so much between the two (although the proper colour encoding in PAL means the 256x192 mono shading hack doesn't work)

erkinalp:
@EtchedPixels: Japan, where, the game console market is prevalent currently, is also NTSC.

PR #245, opened 2015-09-30
coco3: check for pal/ntsc bit in BASIC ROM, set hardware regs accordi…

beretta42:
RF modulated PAL users of coco3 fuzix can enjoy too...


erkinalp:
Next step:Suitable Japanese font to use in NTSC.

PR #246, opened 2015-10-01
tty.c: fix handling of NDELAY behavior...

beretta42:
The way O_NDELAY was handled, caused any reads or writes done under this mode returns EWOULDBLOCK, even after filling the user's buffer with valid data from the inode's buffer.  This fix checks to see if any bytes were read and returns accordingly.


EtchedPixels:
Pushed a different proposal

PR #255, opened 2015-10-07
coco3: easy fixes

beretta42:
easy fixes as suggested by Alan.


beretta42:
ok... did I nail it that time?

PR #256, opened 2015-10-08
Allow DriveWire tty's to by ioctl'd by vt_ioctl().

beretta42:
If we rearrange the order, then the DW tty's will still do tty_ioctl's.


EtchedPixels:
Conflicts with the fixup I did to make it do tty_ioctl.

I'm not sure it makes sense to go to vt_ioctl as things like the keyboard map make no sense.

beretta42:
bah!  OK.. I get it now.  Yeah, I think going directly to tty_ioctl() makes sense.

PR #259, opened 2015-10-17
sh now builds and runs on modern computers.

davidgiven:
This allows sh to build and run on Linux.

(I was debugging a horrible issue elsewhere and found myself needing to step through the sh memory allocation code in gdb. The sh innards are rather scary. And, it turned out, completely correct AFAICT.)

### DIFF
--- a/Applications/V7/cmd/sh/args.c
+++ b/Applications/V7/cmd/sh/args.c
@@ -10,8 +10,7 @@
  *
  */
 
-#include <stdlib.h>
-#include	"defs.h"
+#include "defs.h"
 
 static const char **copyargs(const char *from[], int n);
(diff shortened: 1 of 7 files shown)


erkinalp:
Thanks for porting!

davidgiven:
The sh code is some of the oddest C I've seen for a long time. It'd be really rather interesting to tear it apart some day and put together a retrospective design doc. The memory management structure is... interesting...

Anyway, thanks!

EtchedPixels:
Be warned - Tom Duff tried to for the POSIX shell specification. To quote Tom Duff:

"Nobody really knows what the Bourne shell's grammar is. Even examination of the source code is little help."

PR #262, opened 2015-10-19
MSP430 port

davidgiven:
So here is that MSP430 port.

Apart from the platform code, there are some kernel and library changes needed to support it. These consist of:

- a bunch of alignment fixes --- the MSP430 requires it. (It doesn't trap on alignment errors, it just dereferences the *wrong address*.)
- the indirect queue code we discussed a while back.
- some minor optimisations --- in a direct memory system, we don't need separate functions for u{get,put}{c,w}(), so I've replaced those with macros.

I've checked to make sure this all still builds on 6502test but trs80 and dragon seem broken right now. They do at least get past the kernel build stage. And of course I can't run anything other than MSP430.

Build instructions are in the Documentation/ directory.


EtchedPixels:
Ok .. there are some minor cleanups probably want doing - notably cleaning up the queue macros (which does mean affecting other ports so wants doing after the merge anyway)

Rest looks excellent to me, and wonderfully crazy. I guess putting the kernel in 20bit mode is a later project 8)

EtchedPixels:
Small follow on patches pushed, with those dragon-nx-32 is fine, not checked Z80 yet

davidgiven:
Yup, that's all fine.

Yes, the MSP430 is a wonderfully crazy processor. It's the only *modern* equivalent to the old 8-bit micros I've met. (Modern microcontrollers all have lots of flash and tiny amounts of ram, which means you can't load programs on them.) Incidentally, the FR5969 has 64kB of FRAM --- persistent RAM. Performs like SRAM, but keeps the data when the power goes off. Which means it doesn't distinguish between flash and RAM!

And Fuzix works really well on it. It's fast and snappy and totally doesn't act like something which is consuming less than 2mA running flat out...

More to the point: this is also a fully vanilla gcc-binutils port, with no additional tools or binary converters needed, and will come in really useful when doing the ARM port. Although I will admit than a sane person would probably have done it the other way round.

PR #263, opened 2015-10-19
Minor MSP430 fixes

davidgiven:
This fixes a couple of trivial bugs in your recent changes that were stopping the MSP430 from building.

I also figured out the proper magic numbers. Turns out the test costs 18 bytes of code space...


EtchedPixels:
Why not make config.h also include stdint ? It's not pretty but its allowed 8)

davidgiven:
Do... do you think we *dare*?

Okay, fixed.

PR #268, opened 2015-10-23
Dragon-nx32 graphics fixes

beretta42:
A couple of bugs I noticed while porting this code to coco3.

The endian fix not only allows C structure to be define normally, but also allows the bounds checking code (the other submitted fix) to work (it treats the passed structure as an array of uint16's


EtchedPixels:
Fun... my test cases are buggy 8)

beretta42:
I also noticed that GFXIOC_READ, plops the video data back in the user-space on top of the bitmap header data (len,X,Y,H,W).  I would expect it to fill up the data field (offset 10 ) rather?  This would allow one to grab screen data, then write it back without having to formulate a new buffer?

EtchedPixels:
I think that would be a good change - send patches!

PR #269, opened 2015-10-23
Enable multiprocess mode for the msp430

davidgiven:
As discussed.

Also rearranges the build system a bit --- there were way to many things to change when changing the size of userspace.


EtchedPixels:
All merged - docs are now out of date of course 8)

PR #272, opened 2015-10-24
MSP430 overlay support

davidgiven:
As discussed. This gives me about 2kB of extra user space, as well as enough slack for another slot in the process table (because user space needs to change in 0x200 byte increments). Doesn't seem much, but that's 10% more user RAM.

Copying 2kB at one byte per 8 cycles on a 8MHz device takes 2ms. There are faster memcpys if necessary (it's got a rather nice DMA engine). But I don't think anyone will notice.

It's a shame syscall_fs has to be in common. That would give me another kilobyte...

I am, however, concerned it's rather brittle. If anyone moves a syscall, or adds a new one, my code will load the wrong overlay and Bad Things will happen. And because this will happen on an individual syscall level, nobody will notice until that syscall is exercised.

### DIFF
--- a/Build/platforms/msp430fr5969.ld.h
+++ b/Build/platforms/msp430fr5969.ld.h
@@ -1,8 +1,8 @@
 #ifndef MSP430FR5969_LD_H
 #define MSP430FR5969_LD_H
 
-/* Must be 0x0080 plus a multiple of 0x0200 */
-#define USER_BASE 0xa380
+/* Must be 0x0180 plus a multiple of 0x0200 */
+#define USER_BASE 0x9b80
 #define UDATA_BASE (USER_BASE - 0x0200)
 
 #endif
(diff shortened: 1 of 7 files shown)


EtchedPixels:
The unix_syscall hooking bit is very neat - that's a trick I'd missed that makes it much more workable.

I'm not sure that it needs to be fragile however or that syscall_fs can't in theory be overlaid

For the fragility the syscall to overlay map could be generated by using nm to find which overlay matches which syscall and autogenerate the table perhaps ?

For syscall_fs as there is nothing IRQ related I *think* the following is safe

Save the overlay state in u_data.u_page2 (p->p_page2 is used for swap but it never needs to trash the udata version).
In switchin switch overlays to match the sleep entry state
In fork() clone the overlay state
On entry/exit or somewhere save an overlay state of "userspace", which means "don't care" on the switchin.

and it will then handle the sleeping syscalls

EtchedPixels:
I broke this slightly - I split the syscall files up a bit to make all of them about the same binary size (on Z80 at least). I've also added a tool to generate the array contents of which syscall is in which bank.

davidgiven:
Yup, that's failing to build now. I'll fix it in a couple of days. The tool should be really useful.

PR #277, opened 2015-10-29
CoCo3: color text screens

beretta42:
implemented coco3's 16 color text modes.

### DIFF
--- a/Kernel/platform-coco3/devtty.c
+++ b/Kernel/platform-coco3/devtty.c
@@ -137,10 +137,10 @@ static struct display fmodes[] = {
 };
 
 static struct mode_s mode[5] = {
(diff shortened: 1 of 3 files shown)


EtchedPixels:
Time to add the Atari ST colour escapes to vt.c I wonder
(It supports escape b [code] and escape c [code] to set the fg and bg colour based on the lowest bits of the code value

beretta42:
indeed.   I guess we would just have to add 'b' and 'c' to the big mode
switch in vt.c and do up some macro magic to allow playforms to hook
into... maybe a #ifdef VTCOLOR# in there to call some platform defines.

-brett

EtchedPixels:
Initial patch pushed

PR #280, opened 2015-11-01
Better MSP430 overlay support

davidgiven:
This uses your new tool to figure out which syscalls are in which file; thanks!

All syscall banks can now be used as overlays. Turns out that caching a process' current overlay in udata.u_page2 is unnecessary --- I can figure out which overlay was loaded after switchin by looking at the syscall in udata.u_callno, and I already have code to do that. Total amount of extra code: one subroutine call in switchin. I now have a 27kB userland, which is actually getting quite respectable. (That's about the same amount free as a BBC Micro with tape filesystem and mode 7.)

However, I did have to move unlinki and the guts of _sync into common code, as these functions are called from multiple syscall banks and we don't support cross-bank function calls. (How did this work on hardware banked platforms?)

### DIFF
--- a/Build/platforms/msp430fr5969.ld.h
+++ b/Build/platforms/msp430fr5969.ld.h
@@ -2,8 +2,8 @@
 #define MSP430FR5969_LD_H
 
 /* Must be 0x0180 plus a multiple of 0x0200 */
-#define USER_BASE 0x9b80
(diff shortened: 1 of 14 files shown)


EtchedPixels:
The Z80 banked ports are using an SDCC which is smart enough to have the linker do all the interbank calls and stubs that do call/return bank switching.

PR #313, opened 2016-01-20
Fix alignment and sizing issues

davidgiven:
The only bit of any great interest is the stcpy(), which writes to the user's struct _uzistat structure. I've rearranged it to avoid random member padding upsetting it due to alignment, but I'm not really a fan --- it's very fragile, and any changes would probably fail. I'd much rather import struct _uzistat into the kernel and copy the members one by one, but on small architectures that's probably very expensive.

I suspect the other structures will fail in the same way, but I haven't been bitten by those yet.

### DIFF
--- a/Applications/V7/games/wump.c
+++ b/Applications/V7/games/wump.c
@@ -238,7 +238,7 @@ int main(int argc, char *argv[])
 				if (p->tunn[j] == p->tunn[k])
 					goto init;
 		}
-		qsort(&p->tunn[0], NTUNN, 2, icomp);
+		qsort(&p->tunn[0], NTUNN, sizeof(int), icomp);
 		p++;
 	}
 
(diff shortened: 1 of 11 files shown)


davidgiven:
Blast, I forgot about breaking binary compatibility.

I would much prefer to do a simple member-by-member copy into a structure on the stack, but that's probably too big on small platforms (I suspect SDCC would use ld a, [ix + offset] instructions, which takes eight bytes to copy a single byte; making the entire copy function 272 bytes plus a single uput). 

And we're not using offsetof() because I don't believe the kernel can see the struct _uzistat definition.

I think this is going to need a header file in common between the kernel and userspace.

EtchedPixels:
We can put it in the kernel too 8)

Right now we don't have a common header just copies. That would be a good tidy up to eventually do.

For the moment  can you split the fixes that touch ABI from the ones that don't. An ABI sanity day for 32bit may well be needed but let's just have one of them ?

davidgiven:
Yep, I agree.

davidgiven:
I never said how big the error bars on that number where!

Fixed.

PR #314, opened 2016-01-20
Rewrite of kernel memory allocator

davidgiven:
I suspect nobody is using the kernel memory allocator except for me, because it doesn't work --- kmemaddblk() dereferences p without ever assigning to it. I tried to fix it but was unable to understand the code.

This is a complete rewrite using a very slightly tweaked version of the same algorithm. It's been lightly tested, i.e. it doesn't crash when I use it.

(I haven't seen this algorithm before; it's very neat.)

### DIFF
--- a/Kernel/malloc.c
+++ b/Kernel/malloc.c
@@ -2,8 +2,9 @@
 #include <printf.h>
 
 /*
- *	From Dale Schumacher's public domain dlibs. It's a very simple, very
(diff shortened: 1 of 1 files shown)


EtchedPixels:
I'd pulled it in from dLibs and quickly Ansified it but not tested and debugged it. It's an old algorithm used by a lot of older libraries. I believe it's origin may be the Lyon's book on V6 Unix or even before that.

PR #322, opened 2016-03-03
coco3: Add key repeating for keyboard driver

beretta42:
Added key repeating.  This hack should work on Dragon too.


EtchedPixels:
Added a repeat setting ioctl and pushed to the platform trees. Not yet set a default timeout. I guess that should be done ?

beretta42:
I personally like a .5 sec delay, with .16 sec repeat.   Maybe should express delays in hundredths rather than tenths of seconds?

erkinalp:
Thousandth of seconds maybe? I prefer 0.25s delay plus 0.111s repeat.

EtchedPixels:
Most of our platforms only have 1/50 1/60 or 1/100 clocks at best. We could go to 1/100th but that makes the code much nastier for the conversion

PR #330, opened 2016-03-15
init.c: bugfix init not respawning inittab entries.

beretta42:
https://github.com/EtchedPixels/FUZIX/blob/master/Applications/util/init.c#L197

This line tried to use a initpid entry as a inittab entry.  This causes respawning to go la-la.

### DIFF
--- a/Applications/util/init.c
+++ b/Applications/util/init.c
@@ -173,6 +173,19 @@ static pid_t spawn_process(uint8_t * p, uint8_t wait)
 		return pid;
 }
 
+/* Get condensed inittab entry pointer from index */
+
+static uint8_t *get_inittab( uint8_t index)
+{
+	uint8_t *p = inittab;
+	uint8_t i;
+	for( i = 0; i < index; i++ )
+		p += *p; /* add length to p for next entry pointer */
+	return p;
+}
+
+
+
 /*
  *	Clear any dead processes
  */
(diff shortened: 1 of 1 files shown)


EtchedPixels:
Not convinced this is correct. We still end up copying the wrong bits into ut.ut_id, how about 

````
diff --git a/Applications/util/init.c b/Applications/util/init.c
index 3e80ce1..d233074 100644
--- a/Applications/util/init.c
+++ b/Applications/util/init.c
@@ -180,9 +180,9 @@ static pid_t spawn_process(uint8_t * p, uint8_t wait)
 static int clear_utmp(struct initpid *ip, uint16_t count, pid_t pid)
 {
        uint16_t i;
+       uint8_t *p = inittab;
        for (i = 0; i < count; i++) {
                if (ip->pid == pid) {
-                       uint8_t *p = initpid[i].id;
                        /* Clear the utmp entry */
                        ut.ut_pid = 1;
                        ut.ut_type = INIT_PROCESS;
@@ -199,6 +199,7 @@ static int clear_utmp(struct initpid *ip, uint16_t count, pid_t pid)
                        return 0;
                }
                ip++;
+               p += *p;
        }
        return -1;
 }
````

beretta42:
Oh... I forgot about what happens when the respawn test fails and the loop continues.  Yeah, your fix is better.  Not sure why tty1 utmp entry is screwed up - probably same reason CTTY is never set for the first terminal?

PR #332, opened 2016-03-17
Adventure: makefile for gcc6809

beretta42:
Sorry, some string handling break on gcc6809, so no optimization :( 29k binary


EtchedPixels:
Can you see which file you need to avoid optimizing ?

Alan

beretta42:
Already on my TODO list.

PR #345, opened 2016-05-18
Introducing multicomp09

nealcrook:
Hi Alan,
assuming I've done everything correctly, this is a pull request to add the multicomp09 platform to your repository. I have a second set of changes - typos, makefiles, .gitignore for the repository but they are not specific to multicomp09 so I will present them separately.

I think the code is in reasonable shape but if there are things that you want refined/tidied before accepting it, let me know.

regards,
Neal.

### DIFF
--- a/Kernel/platform-multicomp09/usermem_gime.s
+++ b/Kernel/platform-multicomp09/usermem_gime.s
@@ -1,0 +1,342 @@
+	.module usermem
+
+;;;
+;;;	6809 copy to and from userspace via
+;;;     Color Computer 3 GIME mmu
+;;;
+
+;;; multicomp09 HW registers
+;;; [NAC HACK 2016May03] maybe I should define these all as offsets from IOBASE..
+;;; then index any of them from a single register.
+MMUADR	equ	$ffde
+MMUDAT	equ	$ffdf
+
+;;; bit-fields
+MMUADR_ROMDIS	equ $80		; 0 after reset, 1 when FUZIX boots. Leave at 1.
+MMUADR_TR	equ $40		; 0 after reset, 0 when FUZIX boots. 0=map0, 1=map1
+MMUADR_MMUEN	equ $20		; 0 after reset, 1 when FUZIX boots. Leave at 1.
+MMUADR_NMI	equ $10		; 0 after reset, 0 when FUZIX boots. Do not write 1.
+MMUADR_MAPSEL	equ $0f		; last-written value is UNDEFINED.
+;;; the only two useful values for the upper nibble
+MMU_MAP0	equ	(MMUADR_ROMDIS|MMUADR_MMUEN)
+MMU_MAP1	equ	(MMUADR_ROMDIS|MMUADR_MMUEN|MMUADR_TR)
+
+	include "kernel.def"
+        include "../kernel09.def"
+
+	;; window xfer treshhold - any user/krn space xtfers
+	;; bigger than this will be routed to the banking/windowing
+	;; transfers, rather than the original slower routines
+WINTR	equ	256		; window xfer threshold
+
+	; exported
+	.globl 	__ugetc
+	.globl 	__ugetw
+	.globl 	__uget
+	.globl 	__ugets
+	.globl 	__uputc
+	.globl 	__uputw
+	.globl 	__uput
+	.globl 	__uzero
+
+	; imported
+	.globl 	map_process_always
+	.globl 	map_kernel
+	.globl _krn_mmu_map
+
+	.area 	.common
+
+__ugetc:
+	pshs 	cc	; save IRQ state
+	orcc 	#0x10
+	jsr 	map_process_always
+	ldb 	,x
+	jsr 	map_kernel
+	clra
+	tfr 	d,x
+	puls 	cc,pc	; back and return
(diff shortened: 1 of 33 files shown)


EtchedPixels:
On Wed, 18 May 2016 13:05:01 -0700
Neal Crook <notifications@github.com> wrote:

> Hi Alan,
> assuming I've done everything correctly, this is a pull request to add the multicomp09 platform to your repository. I have a second set of changes - typos, makefiles, .gitignore for the repository but they are not specific to multicomp09 so I will present them separately.
> 
> I think the code is in reasonable shape but if there are things that you want refined/tidied before accepting it, let me know.

Only real issue I have is dev/mbr.c. What stops you putting the MBR at
the start of the media. Not doing that will cause you a world of pain
when moving media to and from a PC class device ?

Alan

nealcrook:
Short answer: it allows me to use a single SD card for multiple OS images. If I put the MBR at the start of the SD card, I would have to devote an SD card to FUZIX *or* do a lot of rework of other software environments.

Long answer: I currently have, on a single SD card, disk images for multiple different OS:
CUBIX
FLEX
NitrOS9
Microsoft 8K BASIC
Buggy debug monitor
CamelForth 
FUZIX
some of these are virtual Floppy images, others are are simply binary blobs. The ROM (CamelForth)
is coded to "know" where each of these is on the SD, and knows the gory details of loading and starting any of them. None of the existing images uses a partition table. I have freedom to put any of the existing images anywhere on the SD, *except* for the CUBIX stuff, which must go at block 0 (because that is the only port I did not do myself, and I would have to recreate it from source in order to make this change). My original plan for FUZIX was to create an ad-hoc boot scheme and hard-wire the disk image offsets into the startup code. The idea I ended up with was to make the boot loader look like an MBR, and contain a partition table - that allowed me to reuse the existing mbr.c -- with the one caveat that I needed to be able to specify a non-zero start address for it.
Even if I could move CUBIX away from block 0 and put the MBR there, I would still have the problem that the other things on the SD do not use a partition table (not an insoluble problem -- I could hide them all in a single partition -- but a bunch of tedious rework)
Since the partition table in the MBR uses addresses that are relative to the MBR, it's possible to dd a chunk of image off the SD starting at the MBR, and treat that chunk as a well-formed disk image. Alternatively, it's possible to dd the fuzix disk images off and modify them using the FUZIX Standalone/ tools.

Neal.

beretta42:
I dont think the other Multicomp OSes do MBR.  I think sharing media is accomplished via hard offsets in that case?

nealcrook:
>> I think sharing media is accomplished via hard offsets in that case?
that's correct. For example, for FLEX I support 4 (virtual) floppy disk images and each of them is at a fixed offset on the SD card, and that offset is coded into the FLEX disk driver. I wrote a PERL script that can create and manipulate those disk images - similar to what lwtools can do for OS9/NitrOS9 disks,
Neal.

EtchedPixels:
In which case can you tweak the submission to look at block 0, and if there is a define also at that offset. That way there is at least a path towards sanity on multicomp and if you ever add a second SD or similar to one you'll be able to handle normal SD images as well.

Alan

nealcrook:
OK, something like this?

````
    do{
        blk_op.nblock = 1;
        if(!blk_op.blkdev->transfer() || le16_to_cpu(br->signature) != MBR_SIGNATURE){
#ifdef CONFIG_MBR_OFFSET
            /* funky disk arrangement. Last change to find MBR */
            if(seen == 0) {
                    blk_op.lba = CONFIG_MBR_OFFSET;
                    continue;
            }
        }
#else
	        break;
        }
#endif
````

**Comment by EtchedPixels** [comment]
Yep but with chance spelled properly - and I think you still need the break if the if fails don't you ?

(I'm fussy about core code, but relaxed about platform code - you own your own problems there ;-))

**Comment by nealcrook** [comment]
I've committed a cleaner mbr.c fix, and a couple of other clean-ups to my platform code. I tested the mbr.c fix in emulation for MBR at block 0 and at block CONFIG_MBR_OFFSET and it worked OK in both of those simple cases.

Neal.

**Comment by EtchedPixels** [comment]
On Thu, 19 May 2016 12:06:13 -0700
Neal Crook <notifications@github.com> wrote:

> I've committed a cleaner mbr.c fix, and a couple of other clean-ups to my platform code. I tested the mbr.c fix in emulation for MBR at block 0 and at block CONFIG_MBR_OFFSET and it worked OK in both of those simple cases.

Merged

Thanks

File: Kernel/platform-multicomp09/usermem_gime.s [hunk 1]
```diff
@@ -0,0 +1,342 @@
...
+	ldd	krn		; get kernel ptr
+	anda	#$3f		; get 16k offset
+	pshs	u		; save data stack
+	tfr	d,u		; U = ptr to window to kernel
+	;; set user bank (at CPU addr 0x4000)
+	lda	usr		; get userspace ptr
+	rola			; rotate top three bits
+	rola			; down to the bottom of A
+	rola			;
+	anda	#3		; mask off junk [NAC HACK 2016May03] why not 7??
```
beretta42: rola rolls the bits through C first.

PR #348, opened 2016-05-24
Add termcap and curses to 6809 libraries; fix compilation fails

nealcrook:
In trying to get some code to compile I found that Libraries/libs/Makefile.6809 was not at parity with Libraries/libs/Makefile. When I added the missing components, I got compilation errors due to conflicts between the function prototypes in the .h files and the definitions in the .c files. Is the z80 compiler so un-fussy? I *think* my fixes are correct.
Also, added NAME_MAX to dirent.h after consulting the contents of the "grown up" version on my Linux system.


EtchedPixels:
SDCC is fussier about some things than gcc but slacker about others. In particular it's not fussy about const matching on prototypes.

PR #354, opened 2016-05-31
Apps3 - a working man

nealcrook:
Replace util/man.c with a different GPL'd version from ELKS, suitably masaged to run under FUZIX. Unlike the old version, this one works without an external nroff (which we didn't have). man.1 as proof
of concept.
ucp-extras-script.txt creates the man/man* directories and inserts man.1
Fixed util/termcap.c so that it compiles on 6809
Added all the new executables created by Tormod's updates to ucp-extras-script.txt

### DIFF
--- a/Applications/V7/.gitignore
+++ b/Applications/V7/.gitignore
@@ -1,11 +1,32 @@
 # from cmd/
 ac
+accton
+at
+atrun
 col
+comm
+cron
+crypt
 dc
+dd
+deroff
 diff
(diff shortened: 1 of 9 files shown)


tormodvolden:
BTW, I have sent patches to Alan fixing ucp to accept comments :)

nealcrook:
Tormod - that's great. I'm glad I only spent a few minutes thinking about what that would involve!!

PR #360, opened 2016-06-16
Add dw_transaction ioctl for Drivewire

beretta42:


EtchedPixels:
The ioctl should ideally be uniquely numbered. Need to think about how to code device specific ioctls tonight. I shall sleep on it 8)

beretta42:
Ah..  I was afraid of that.  So ioctl numbers should be unique across
different kind of devices?  btw, I think Tormod and Neal are ok with the
dw_transaction ioctl rather than the higher level mount call.

Brett M Gordon

EtchedPixels:
Updated kernel.h - use 0x0500-0x050F for drivewire ioctls (for those that should only be usable by root then 0x4500-0x450F). I think your ioctl falls into the latter category otherwise a user could use it to patch the root file system and edit /etc/passwd etc 8)

PR #361, opened 2016-06-17
bank16k:  bug fix pagemap_realloc

beretta42:
bug: released map pages in udata being filled in with simultaneously freed common page 3.  On context switch back to user-space, fuzix would explode because common has been moved (just like the case of want>have) Fix: fill in freed elements with our new "top" page, and proceed to copy our common code.  This also echos what pagemap_alloc() does.

### DIFF
--- a/Kernel/bank16k.c
+++ b/Kernel/bank16k.c
@@ -124,10 +124,11 @@ int pagemap_realloc(usize_t size)
 	if (have > want) {
 		for (i = want; i < have; i++) {
 			pfree[pfptr++] = ptr[i];
-			ptr[i] = ptr[3];
+			ptr[i] = ptr[ want - 1 ];
 		}
 		udata.u_ptab->p_page = udata.u_page;
 		udata.u_ptab->p_page2 = udata.u_page2;
+		program_vectors(&udata.u_page);
 		return 0;
 	}
 


EtchedPixels:
Ok so the vectors need fixing in the case we drop to a single 16K page on some platforms because we trashed low memory.

The other change looks wrong. Whatever is going on ptr[3] holds the upper page that we want to distribute and keep. If we have 0,1,2,3 and we trim to 16K we want to end up with 3,3,3,3.

beretta42:
The algo as it stands would try to free the common page in a have=4 and want=2 case no?

EtchedPixels:
In that case we would have 0 1 2 3 originally and we would do

have > want (true)
for i .. 2 ..3
     free 2 3

So yes.. but I think the bug is actually that we should do

for (i = want ; i < have ; i ++)
{
   pfree[pfptr++] = ptr[i-1];
   ptr[i-1] = ptr[3]
}

Then we will free pages 1 and 2 while setting both to page 3 (the existing common) which was the original intent. I suspect the bank16k-low.c code is also wrong.

The program_vectors fix is still needed because if you go to want = 1 we put page 3 over page 0 and page 0 in some systems contains vectors as well but we should test if (want == 1) because it's expensive on some platforms.

Make sense or am I still missing something ?

beretta42:
Nope.. your right on all of the above.

EtchedPixels:
Fix pushed I hope - please review the revised logic

beretta42:
I see you turned off the interrupts in all have/want cases...that's
probably a good  thing.  Looks good to me; I'll try it out on some chmem'd
bins, and see how the changes work in real life.

PR #362, opened 2016-06-18
Tweaks to allow library to build on 6809

nealcrook:
In termcap.h there was an extra closing brace on tputs
In tzset.c there was a VOID that caused unhappiness; changed to void. Also expunged DOS line endings and trailing whitespace.
Now builds and apparently working on 6809.


EtchedPixels:
Thanks

PR #363, opened 2016-06-18
First exposure of tool for per-application control of root disk contents

nealcrook:
Script is Standalone/filesystem-src/build-filesystem-ng
fuzixBLAH.pkg files are distributed
Documentation/build-filesystem-ng.txt describes current functionality and future proposed functionality.
Can be used to create working disk images so I'm publishing it to get feedback before investing more effort down a possibly flawed path,
Neal.


tormodvolden:
It is a cool concept with packages and everything. I like perl sometimes, but wouldn't it be possible to make a simpler file format (maybe just whitespace instead of colons) that a simple sh script can deal with? One could argue that with this extensive format it would be marginally longer to just write out cp a b; chmod b most places :)

Also, the existing makefiles already maintain a list of files (for the size.report target), maybe we could use this for the "easy" executables. Sure, text files and such would need extra support.

Just brainstorming here, one idea could be instead of a table, a more declaration-driven(?) approach:

files /usr/bin      # declares destination for following entries (will also be created)
mode 755      # declares mode
ls      # single word is simple entry
subdir/tool   # pathnames will be removed
include filelist    # the file filelist could be a make target, like size.report, with many entries
files /usr/share/prog
mode 644
readme.txt

levee 755 /usr/bin   # 3 words is all in one

mode 60660
nodes /dev   # following entries are maj/min - name
1025 mem   # (not sure which order is best)
1026 zero 20444  # 3 words are all in one in "nodes" mode?

beretta42:
looks groovy. I like the "metapackages".  Could this package system somehow be (re)used at run-time? Would it help if the standalone version was in C?  Maybe export it's built database to fuzix?

nealcrook:
I thought about whether there was any reason for this tool to run under FUZIX and I could not think of one, so I chose PERL because I am at least 10x more productive programming in PERL than in C.

It would be trivial to remove the ":" separator and just use whitespace, I like that idea.

The reason why I chose a separate file rather than trying to spit something out of Make was that we already have per-CPU makefiles so that is 2x the maintenance over this "package file" format. The downside, of course, is yet another file/yet another syntax.

>> cp a b; chmod b
maybe that would be a good idea; just allow multi-line UCP commands and translate them, and also
take care of being-in-the-right-directory.

>>Maybe export it's built database to fuzix
it might be nice to generate /etc/fuzixfs.pkg on the target system to list the component packages. It can itself be in pkg format so that it could act as a metapackage in order to (re-)generate an updated/identical system,

>>Could this package system somehow be (re)used at run-time
the use-case would be if you had a running fuzix system and you wanted to add/remove packages - rather than the current situation where (I, at least) just build everything that's available. I think that the .pkg format would be adequate to describe a package bundle that could be created as (eg) a .tar file on the host system and slopped over to the target somehow for "installation"

erkinalp:
Some source code files are not associated with *any* package.

nealcrook:
> Some source code files are not associated with any package.

The scope of this code is as a way of generating a root/system disk for a target. The package files only list stuff that should end up on the target machine. Therefore, the source code files should not be listed in the package files (there are exceptions.. for fforth, for example, there is a forth program that can be run under the control of fforth on the target machine: this does appear in the package file, and ends up in some place like /usr/src/fforth. See fuzix-fforth.pkg)

erkinalp:
I see, `factor.c` ported from toybox is orphan. Should I file a PR for it?

nealcrook:
> Should I file a PR for it?

I have added it to the appropriate package file. If this push request gets rejected, I will go back and add it to ucp-script.txt

beretta42:
1+ for passing through ucp commands.  My custom build files also create links (vi --> levee).  Passing through commands to ucp would also allow for names pipes, etc...  

as for my goofy run-time package ideas:  Hmmm.. my first couple linux's where slakware... about 10 3.5" floppies, and a pipe. ;)  But, from my studies I find .deb and .rpm are pretty much just that: a nice database record file (like your pkg files) plopped into a tar ball with the associated files.

EtchedPixels:
Definite -1 to pass through commands because it means we then can't use the same rules for packages installed by other means. If need be add to ucp

tormodvolden:
> The reason why I chose a separate file rather than trying to spit something out of Make was that we already have per-CPU makefiles so that is 2x the maintenance over this "package file" format. The downside, of course, is yet another file/yet another syntax.

We already have two sets of Makefiles (plus .gitignore...) where we need to maintain the list of binaries. If we don't make use of the Makefiles here, it will be yet another place to keep up to date. Maybe I need to clarify that I don't want to stuff all this "package" logic into the Makefiles, only that the Makefiles spit out a list of binaries, that the "package" logic can use, instead of having its own list.

On the topic of two (and growing?) sets of Makefiles I am a little sad that David's build system is bit-rotting. It could maybe provided a more unified maintenance. The disadvantage of that system is that it is so complex that mere mortals totally lose track of what is going on.

> 
>     cp a b; chmod b
> 
> maybe that would be a good idea

Or an "install" command. Maybe ucp should get one? E.g:
 install -m 644 fforth_tests.fth /usr/share/forth
It could alternatively be seen as a table with -m as separator :)
And it might be convenient that it, unlike the "real" install, would implicitly create the destination directory (but not allow a different destination file name).

EtchedPixels:
I am still slowly trying to move all the Z80 non-kernel code to use "fcc" and I still have making the linker accept ".o" as a valid extension for .rel files on my TODO list. At that point maybe we have more chance to unify the build system and sort it out properly.

nealcrook:
> I am a little sad that David's build system is bit-rotting.

I agree. I would love to see a write-up on how it works and how to generate new .mk files.

In case it wasn't clear, the existing command-set (File Node Directory) was not intended to be complete, but rather the minimum set needed in order to reproduce the existing capability. I can see that the ability to generate soft links would be useful.. and I'd be interested to see an example that calls for the creation of a named pipe.

Regarding using the makefile to create a file list.. If there's any makefile involvement, I think it would be cleaner to have the whole package file generated.  Hand-written and auto-generated package files can co-exist (even in the same directory) and all or any of the hand-written ones could be migrated to auto-generated form in the future, if the need/will existed.

Based on the feedback so far I propose these changes:

1/ switch from : to white=space as a separator (update code and all package files and documentation)
2/ add "Link" command to the Futures list

beretta42:
Maybe add a chmem command to ucp/packages too.    That way we bank16ers can restrict memory allocations from "all-space" executables.  

BTW:  almost had David's build system working here, but ucp was broken/crashing at the time. (Not that i understood half of what i was looking at.)

nealcrook:
>  I propose these changes..

Committed.

PR #365, opened 2016-06-23
Dwtime

beretta42:
A drivewire RTC without too many shenanigans.  Should should allow for setting system time different than the dw server - provides elasped time for keeping the system time in-sync, anyway.


EtchedPixels:
Ok I don't understand why dwtime can't be in user space ?

EtchedPixels:
It would be useful to have a userspace version for boxes like the trs80 that have a clock source of their own so don't want/need the overhead yes. Mind you I still need to debug the trs80 drive wire - too many emulators not enough time

beretta42:
I'll put it on my TODO list.  We probably should scrounge up a "date"
program that does setting of the system clock too.

PR #375, opened 2016-07-12
6809: compile libs separately

beretta42:
This matches what is being done on the z80 makefiles.

### DIFF
--- a/Applications/util/Makefile.6809
+++ b/Applications/util/Makefile.6809
@@ -108,7 +108,7 @@ SRCS  = \
 	write.c \
 	xargs.c
 
-SRCSM = tget.c
+SRCTC = tget.c
 
 SKIPPED = \
 	tiddles.c \
@@ -116,13 +116,13 @@ SKIPPED = \
 
 OBJS = $(SRCS:.c=.o)
 OBJSNS = $(SRCSNS:.c=.o)
-OBJM = $(SRCSM:.c=.o)
+OBJTC = $(SRCTC:.c=.o)
 
 APPS = $(OBJS:.o=)
 APPSNS = $(OBJSNS:.o=)
-APPM = $(OBJM:.o=)
+APPTC = $(OBJTC:.o=)
 
-all: $(APPS) $(APPSNS) $(APPM) size.report
+all: $(APPS) $(APPSNS) $(APPTC) size.report
 
 $(APPS): $(CRT0)
 $(APPSNS): $(CRT0NS)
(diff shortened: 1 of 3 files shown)


nealcrook:
Hoorah. I wondered who would crack first and do this!
I also have the stdlib.h mod. in my merge request.
Did this fix your levee build contention? I am hacking in levee right now, and have disambiguated the calls that conflict with standard library stuff. Now working through some other issues to get it to work cleanly on my platform

beretta42:
It definitely helps levee compile, but I think I still need to compile it
with "-O0".  The only other mod I do is increase the buffer size to 8192,
for editing a few bigger files.

Hmm.. if you're patch does the same labs() fix, I wonder if git will merge
mine?  Oh well.. It's not hard to cherry-pick a new PR if needs be.

Maybe try my patch and see if it works OK for you?  I'm absolutely abysmal
at noodling with makefiles.

-Brett

nealcrook:
Hi Brett,
I copied your makefile changes to my messing-with-levee branch and they make things work much, much better. I now have a vanilla "ANSI" build running more-or-less OK. A couple of questions on your levee experience though:

1/ when you start up levee, do you get a startup banner: version and author? I suspect not. It gets written to line 1 then immediately overwritten.

2/ can you edit an existing file, or read a file into an existing buffer? For me I always get a message that the file size is "-1 bytes" which looks like an untested "fail".

3/ do you get any weirdness with screen redraws? I get whole unused lines filled with "~" (instead of
just a single ~ in the first column)

I am part-way through converting lv.doc to a man page (helped by grabbing the man page from the latest version of levee)

I also spotted that the levee command jump-table is 256 bytes but it can be 128 bytes (forbid top-bit set) for an instant saving of 128 bytes.

beretta42:
On Jul 13, 2016 4:48 PM, "Neal Crook" <notifications@github.com> wrote:
>
> 1/ when you start up levee, do you get a startup banner: version and
author? I suspect not. It gets written to line 1 then immediately
overwritten.

Yes Sir. Coco3 does same. (Some while ago it would pop up on bottom line.)

>
> 2/ can you edit an existing file, or read a file into an existing buffer?
For me I always get a message that the file size is "-1 bytes" which looks
like an untested "fail".

definitely had this error - i think compiling with "-O0"  fixes this.

>
> 3/ do you get any weirdness with screen redraws? I get whole unused lines
filled with "~" (instead of
> just a single ~ in the first column)

Nope.. I dont get that.

> I am part-way through converting lv.doc to a man page (helped by grabbing
the man page from the latest version of levee)
>
> I also spotted that the levee command jump-table is 256 bytes but it can
be 128 bytes (forbid top-bit set) for an instant saving of 128 bytes.

PR #379, opened 2016-07-16
Changes for existing commands, add "less"

vandys:
Add an early "less" version, fits in ~40k.
Fix touch(1) to properly display errors.
Modify filesystem creat/open code to return specific errors.
Add /etc/group.
Have prototype root filesystem created with 755 protection on /
Teach "size" in Standalone about 6809 exec format; try to keep Z-80
 support by keying off magic number.

(Alan, not sure what granularity of pull request you prefer, this one's a bit bigger than
 previous ones.  Just let me know!  --Andy)


EtchedPixels:
Its both the granularity and the mix of stuff in a single commit.

One thing one commit please so

- One for adding less
- One for stdint.h
- One for touch
- One for /etc/group
- One for the standalone tools

For the kernel side changes some more detailed explanation is needed as well

PR #384, opened 2016-07-26
Add a schlocky ashrsi3 to 6809, fix devdw.

beretta42:
This needs some optimization (what's the easiest way to propagate the sign bit when optimizing?).  ashrsi3 follows same calling convention as existing ashlsi3.  This makes new devio changes compile on 6809.  Plus simple bug fix for devdw.


EtchedPixels:
The optimisation worth spotting for the kernel is >> 8 and >> 9 as we only ever generate the cases
7 8 and 9 (and I think 7 isn't used on 6809 boxes as none of ours have 128byte/sector FM discs)

EtchedPixels:
As we know it will be +ve we could also cast to unsigned before the shift and implement __lshrsi3 instead so the sign problem goes away ?

PR #393, opened 2016-08-07
zx128: Betadisk-128 driver and filesystem building definition

atsidaev:
Betadisk-128 is possibly the most popular ZX Spectrum floppy controller in the world (taking into account its huge popularity in the Eastern Europe). But it hardly can be used directly because WD1793 (which it is based on) ports are only available from the TR-DOS ROM. 

But while implementation of "true" driver is tricky, the simple primitives for the heads positioning or 256 bytes sectors read/write may be executed from that ROM using simple hacks.

Bad thing here is that these hacks depend on TR-DOS firmware version. Currently everything is made for the 5.03, which is most widespread in ex-USSR (and seems to be the latest official version released by Techology Research Ltd). For other versions or earlier Betadisk-48 controller some more generic solution is needed.

### DIFF
--- a/Build/platforms/zx128.mk
+++ b/Build/platforms/zx128.mk
@@ -1,0 +1,50 @@
+$(call find-makefile)
+
+O = rel
+A = lib
+
+# CPU architecture and which syscall generator to use.
+
+ARCH = z80
+SYSCALL_GENERATOR = syscall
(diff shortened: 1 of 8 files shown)


EtchedPixels:
Thanks

PR #394, opened 2016-08-07
Standalone tool changes--root filesystem protection and size command

vandys:
Default root of a created filesystem to 755, not 777.
Add endian support so "size" can look at 6809 executables.


EtchedPixels:
No to the mkfs one - 777 is expected there
Rest looks good though

EtchedPixels:
Changes to size merged, forgot about this sorry

PR #403, opened 2016-08-28
dwdate: sets system time from drivewire via userspace

beretta42:
good for once off'ing during boot from an rc file, or cron'ing.

### DIFF
--- a/Applications/dw/Makefile.6809
+++ b/Applications/dw/Makefile.6809
@@ -15,7 +15,7 @@ CRT0 = ../../Library/libs/crt0_6809.o
 .SUFFIXES: .c .o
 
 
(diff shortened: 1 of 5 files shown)


EtchedPixels:
stime() sets the system clock so shouldn't have a type field added. It's a standard Unix API so if we need a version with a type it would need to become _stime() with a wrapper in libc.

beretta42:
Libc is already wrapping _stime() to become a normal stime(). Not that the type param does anything anyway ;)  We could simplify things and get rid of "type".  The Libc wrapper might not be needed then?

EtchedPixels:
I'd forgotten I did that when I did _time. Makes sense as it's very cheap to then add other time setting types.

PR #409, opened 2016-08-30
termios.h: stty didn't like winsize not being defined

beretta42:
I need to do this to build system ?  or should winsize be local to stty.c ?


EtchedPixels:
stty needs fixing - patch pushed

PR #419, opened 2016-09-01
makefile: make prerequisites for new makeversion.

beretta42:
This seems to force makeversion and friends before decending into platform code


dfffffff:
Yep that fix the issue

dfffffff:
I think a dependency is still missing, first time a tried to compile the kernel after applying your fix I got this:

usermem_std-6809.s:3:31: lowlevel-6809.s:62:31:error:  platform/kernel.def: No such file or directory
error: platform/kernel.def: No such file or directory

tried a second time, the error was gone.

Issue #420, opened 2016-09-02
make fail to build the kernel on parallel build


dfffffff:
make fail to build the kernel on parallel build (was working before).

this fail:
```shell
make -s clean && make -sj4
```
but this is ok:
```shell
make -s clean && make -sj1
```

I get this error:
```
make: *** [lowlevel-6809.o] Error 1
make: *** Waiting for unfinished jobs....
usermem_std-6809.s:3:31: error: platform/kernel.def: No such file or directory
make: *** [usermem_std-6809.o] Error 1
```

dfffffff:
now ok if I add to Kernel/Makefile
```makefile
$(OBJS): target
```

it failled because in image.mk we have:
```
fuzix.bin: target $(OBJS) ...
```

dfffffff:
I think I was lucky that it worked before on parallel build

PR #421, opened 2016-09-02
fix Kernel parallel build

dfffffff:
fix Kernel parallel build
issue #420


EtchedPixels:
Merged but please create separate pull requests for different things
Also if you re indent stuff please commit the re-indent then the changes so it's easier to review

Thanks

dfffffff:
Sorry about for the which commit, it was not supposed to be pulled with the makefile fix.

PR #426, opened 2016-09-15
tchelp: fixes for 6809 platforms

beretta42:


EtchedPixels:
tchelp is a helper so it doesn't go in /bin but belongs somewhere like /usr/lib

beretta42:
PR updated...

PR #431, opened 2016-09-22
coco3: remove drivewire vwins: not worth memory usage in level 2

beretta42:
These use too much memory in Level2, and not worth their "gee-whiz" factor.

### DIFF
--- a/Kernel/platform-coco3/config.h
+++ b/Kernel/platform-coco3/config.h
@@ -70,7 +70,7 @@ extern unsigned char vt_map( unsigned char c );
 #define CMDLINE	0x88	  /* Location of root dev name */
 
 /* Device parameters */
-#define NUM_DEV_TTY 10
+#define NUM_DEV_TTY 6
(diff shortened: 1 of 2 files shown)


EtchedPixels:
If you look at the tty queue functions you can replace those with your own and put the tty buffers in a private memory page. David actually did that at one point for the msp340 which is why we have GETQ() and PUTQ() that you can override with calls to a private asm routine in common that does something like

bank in tty page
read/store byte
bank in old page

and you need to set the pointers correctly for the other page mapping in your devtty.c

beretta42:
ah... cool.  I didn't catch that way-back-when.  Well, that'll save
boat-loads of kernel memory. Thanks Alan.

My vt code is pretty lousy, btw.  I was thinking of replacing the slow
stuff with ASM, if that's cool with you.

-Brett

EtchedPixels:
Makes sense for all the low level bits anyway

PR #435, opened 2016-09-24
cut: fix output

erkinalp:
replace printf() with just a string by a puts()
no need to go on formatting just to print unmodified string


erkinalp:
Apologies for inactivity.

erkinalp:
I know about gcc, yet there is a magic in that optimisation. puts() normally adds LF at the end. However that form is not allowed to have LF at the end.

PR #436, opened 2016-09-25
fixed compile error: added ssize_t, changed size_t to unsigned

dfffffff:
There is a reason why size_t was signed ?


erkinalp:
I cannot think, `size_t` must be unsigned, `off_t` and `ptrdiff_t` must be signed.

EtchedPixels:
size_t unsigned, ssize_t signed

PR #439, opened 2016-10-05
coco3: bank in vt code

beretta42:
This creates a banking system for the vt.c (and lower) code.  Special note of interest:  from the bank code's pov: external linkages are awked from main kernel map file, and exported linkages are set up as a table of pointers.  And nothing shows you how poorly you've factored your code quite link banking it :)

### DIFF
--- a/Kernel/platform-coco3/Makefile
+++ b/Kernel/platform-coco3/Makefile
@@ -1,5 +1,5 @@
 
-CSRCS = ttydw.c mbr.c dwtime.c bank0.c
+CSRCS = ttydw.c mbr.c dwtime.c
 CSRCS += devices.c main.c libc.c devsdc.c
 
 CDSRCS = ../dev/devide_discard.c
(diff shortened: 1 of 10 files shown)


beretta42:
it was all fun and games until one of the external functions called into code that was banked over:

```
call stack:
   entry        0100    crt0.s
   start        7094    crt0.s
   main()       70ac    main.c
   initialize() 51f5    main.c
   initcon()    6ee4    unixcall.c
   tcgetattr()  1087    tcgetattr.c (libc)
   ioctl()      114d    
   -- kernel --
   __ioctl      0ecb    syscall_fs.c
   _d_ioctl     5482    devio.c
   gfx_ioctl    8a35    devtty.o
   vt_ioctl     8c51    video_link.o
   -- bank0 --
   vt_ioctl     233c    vt.c
   tty_ioctl    8141    tty.c
   jobcontrol_ioctl 323C  level2.c  <- kaboom !
```

beretta42:
ahh... from /FUZIX/Kernel/Makefile:
#       Assume we'll never be doing overlays with level2 or networking

doh!

EtchedPixels:
The low level video could should be bankable but as you observe not the higher level. Given you've got an MMU than can work in 8K sections and your stacks and common are in the top 8K you could just bite the bullet and map a kernel page from C000-DFFF perhaps ?

For banking you want to make the linker do it. The Spectrum 128K port runs the entire kernel code space in a 16K wide window and the linker does most of the work for this. It's not entirely linker driven yet as I've not taught the linker to decide what code goes in which bank but what it does is a mix of compiler and linker ops that make the rest transparent.

Every function call goes from being

                  call _foo

to
                  push af
                  call foo
                  pop af

and with the arguments offset by 2 bytes more

and the linker replaces these for any inter bank calls with

                               call _ bank_n_to_m
                               .word address

and the bank to bank code stacks the old bank (in the space that used to have the dummy af), does the switches, switches back on return.

For relocations (function pointers) it generates a stub routine that does the above and uses that address.

With that quite fancy banking is doable but by hand it's just a nightmare.

Alan

beretta42:
Ok.. I'm with you on this.  I did something *similar * a long time ago on the coco2.  I ended up with a specialized linker for asxxxx to produce binaries for a run-time linker.  I check with William (lwtools)  for advise.

PR #442, opened 2016-10-09
ps.c: fix parent pid calculation

beretta42:
sdcc pointer magic?

### DIFF
--- a/Applications/util/ps.c
+++ b/Applications/util/ps.c
@@ -68,7 +68,7 @@ int do_ps(void)
             close(pfd);
             return 1;
         }
-        ppid_slot[i] = ptab[i].p_tab.p_pptr - ptab[0].p_tab.p_pptr;
+        ppid_slot[i] = ((unsigned)ptab[i].p_tab.p_pptr - (unsigned)ptab[0].p_tab.p_pptr)/sizeof(struct p_tab);
     }
     close(pfd);
 


erkinalp:
Parent PID *computation*.

EtchedPixels:
Original looks correct to me, and your patch breaks it if the platform is 32bit ptr/16bit integer

In proc.h p_pptr is a struct p_tab *, so this is struct p_tab *  - struct p_tab * which yields a numeric value scaled by the size of the structure.

What are you seeing ?

beretta42:
cool. I guessed gcc would scale the difference.  It doesn't appear that is what is happening.  The pointers are correct given my platform, but the math seems to get all screwed up:

```
diff --git a/Applications/util/ps.c b/Applications/util/ps.c
index 9ebcb51..b4fbb18 100644
--- a/Applications/util/ps.c
+++ b/Applications/util/ps.c
@@ -69,6 +69,7 @@ int do_ps(void)
             return 1;
         }
         ppid_slot[i] = ptab[i].p_tab.p_pptr - ptab[0].p_tab.p_pptr;
+	printf("i= %d, slot= %x, p_pptr = %x, ptab[0] = %x\n", i, ppid_slot[i], ptab[i].p_tab.p_pptr, ptab[0].p_tab.p_pptr );
     }
     close(pfd);
 
```

and the output:

```
 # ps
i= 0, slot= 0, p_pptr = c139, ptab[0] = c139
i= 1, slot= 0, p_pptr = c139, ptab[0] = c139
i= 2, slot= 0, p_pptr = c139, ptab[0] = c139
i= 3, slot= af8c, p_pptr = c181, ptab[0] = c139
i= 4, slot= b7c1, p_pptr = 0, ptab[0] = c139
i= 5, slot= b7c1, p_pptr = 0, ptab[0] = c139
i= 6, slot= b7c1, p_pptr = 0, ptab[0] = c139
i= 7, slot= b7c1, p_pptr = 0, ptab[0] = c139
i= 8, slot= b7c1, p_pptr = 0, ptab[0] = c139
i= 9, slot= b7c1, p_pptr = 0, ptab[0] = c139
i= 10, slot= b7c1, p_pptr = 0, ptab[0] = c139
i= 11, slot= b7c1, p_pptr = 0, ptab[0] = c139
i= 12, slot= b7c1, p_pptr = 0, ptab[0] = c139
i= 13, slot= b7c1, p_pptr = 0, ptab[0] = c139
i= 14, slot= b7c1, p_pptr = 0, ptab[0] = c139
USER	  PID	 PPID	STAT	WCHAN	ALARM	COMMAND
root	    1	    1	Asleep	c139	0    	/init
root	   12	    1	Ready	0000	0    	-sh
root	   13	    1	Asleep	9f4e	0    	/init
root	   14	    0	Running	0000	0    	ps
# 
```

EtchedPixels:
Looks like a compiler bug to me at first glance - does it break with other compiler builds ?

dfffffff:
seems ok with gcc 4.3.6, can you try with -O0 ?

beretta42:
just had a change to test with -O0 (and other compiler options)  nothing seems to make ps work for gcc4.6.4.  Looks like I'll have to take a look at the .S files.

PR #443, opened 2016-10-11
sync: gcc complains about sync being void

beretta42:
is it possible I haven't built the userspace in 17 days to catch this ?!?!


dfffffff:
main() should return something
```C
int main(int argc, const char *argv[])
{
    sync();
    return 0;
}
```

beretta42:
yeah, probably should.

PR #451, opened 2016-10-30
ue: add coco3 link script

beretta42:


EtchedPixels:
Seems odd - it should be finding the one in ../util ?

beretta42:
On closer inspection of the makefile, it really should.  Maybe i was seeing things earlier.  Lemme try again.

beretta42:
yup.. I was seeing things.  nevermind.

PR #455, opened 2016-11-08
6809: improve compatibility with other assembler

dfffffff:


dfffffff:
That imply nothing for the future, other assemblers may need to adapt.
The other issue was not the label length but the lack of white space between the label and the opcode.

beretta42:
Ah..ok..I see.  Yeah, that's weird not to have a space.  Local labels are not terribly standard in the common assemblers either.

PR #456, opened 2016-11-13
6809: syscall number is 8 bit

dfffffff:
Syscall number are 8 bit, so the high byte (A reg) is always zero.
Saves one byte of memory by syscall.
ABI compatibility with previous code is preserved.

### DIFF
--- a/Kernel/lowlevel-6809.s
+++ b/Kernel/lowlevel-6809.s
@@ -93,7 +93,7 @@ unix_syscall_entry:
 	SAM_KERNEL
 	std ,y++
 	SAM_USER
-	ldd 1,s		; stacked D register -> syscall number
+	ldb 2,s		; stacked B register -> syscall number
 	SAM_KERNEL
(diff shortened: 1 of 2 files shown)


EtchedPixels:
For the sake of one byte I'd prefer not to burn any bridges

dfffffff:
If you look at the code only B register is saved, why use a 16 bit then.
```
 	SAM_USER
	ldd 1,s		; stacked D register -> syscall number
 	SAM_KERNEL
 	stb U_DATA__U_CALLNO
```
This is not a good idea to take less memory and cpu cycle?
Please explain why this could burn bridge, I'm missing something?

EtchedPixels:
Because right now the userspace passes 16 valid bits. So in future if we need to expand something or add flag bits we can use the other 8bits safely. If we take that out then binaries will pass a random 8bits so we can never re-use them.

dfffffff:
So this is for forward compatibility, a good thing then.

PR #465, opened 2016-11-22
Application BCPL updates for all platforms.

nealcrook:
Add make files for 6809 and linux (host)
Add minimum set of files from BCPL kit
Modify icint.v to support compile-time generation of a variant that implements file-based demand-paged virtual memory for the INTCODE virtual machine, so that memory restricted FUZIX targets can run the
compiler natively.
Add script (the first FUZIX shell script?) 'bcpl' to automate the compilation process
Expand README.
Aside: fishing around in the loft today for some 1979-era copies of Electronics Today International to sent to an archivist in California, I came across the 8" floppy that contains the BCPL programs that I wrote on Southampton University 6809 systems in 1983.


EtchedPixels:
Seems to have a conflict on README so I'll fix that up

For the game engine (games/qrun..c) that did virtual memory I used least regularly used (which tends to perform better but in hardware is harder to implement), and I also kept some pointers to speed up checks for 'next word' and for 'same bank'. Might be worth doing some of that for the BCPL cintcode eventually - especially for the instruction fetches ?

Looks good

EtchedPixels:
Actually can I get you to fix up the BCPL README conflict because I'd like to be sure it's fixed up right

nealcrook:
Grr. This branch has ended up in a nasty state. Something to do with the fact that I accidentally merged the early parts of these changes into my sh branch and then reverted them there.. Those reverted changes do not now appear as part of this merge (even though they are visible in my branch). The affected files are (1) the README (leading to the conflict) and (2) a set of new files I added (from BCPL kit and some example programs). I'm outside my git comfort zone here but I think my best solution is to rebase this set of changes and then re-commit the missing files. Does that sound sane?

nealcrook:
OK, I think that might be fixed now. Feeling nostalgic for CVS today.

EtchedPixels:
You can always solve it CVS style and just create a new branch and add patches to that based on the messed up one. I do that a fair bit/

Also really useful is either quilt or stgit which let you keep a stack of patches on top of  a git tree.

PR #472, opened 2017-01-04
Applications: levee: Build versions for ansi, vt52 and termcap

willsowerbutts:
The termcap version needs work still, should load from /etc/termcap rather than $TERMCAP.


EtchedPixels:
Actually termcap proper favours $TERMCAP over /etc/termcap so that you can set $TERMCAP once at login and then avoid the search of the termcap file.

PR #473, opened 2017-01-06
Kernel: bread() should mark blocks as free on read failure

willsowerbutts:


EtchedPixels:
Want to think about that and if we actually need a BF_BUSTED type status but yes something needs doing

willsowerbutts:
BF_BUSTED feels pointless to me -- it precludes a later re-reading of the sector which might succeed (eg if the floppy disk is swapped for another), and it ties up a (scarce) entry in the bufpool forever.

EtchedPixels:
Yes - agreed. I wanted to trace through the consequences first

PR #474, opened 2017-01-06
New features

willsowerbutts:
Alan -- this series bundles up a number of new features I have been working on.

### DIFF
--- a/Applications/flashrom/Makefile
+++ b/Applications/flashrom/Makefile
@@ -1,0 +1,14 @@
+FCC = ../../Library/tools/fcc
+FCCOPTS = -O2
+
+.SUFFIXES: .c .rel
+
+LIBS = ../../Library/libs/syslib.lib
+
+all: flashrom
+
+flashrom:	flashrom.c
+	$(FCC) $(FCCOPTS) flashrom.c -o flashrom
+
+clean:
+	rm -f flashrom core *~ *.asm *.lst *.sym *.map *.rst *.lk *.noi *.rel *.bin
--- a/Applications/flashrom/flashrom.c
+++ b/Applications/flashrom/flashrom.c
@@ -1,0 +1,551 @@
+/* 
+    FLASHROM: in-system flash ROM programmer for FUZIX
+    (c) Will Sowerbutts <will@sowerbutts.com> 2017-01-02
+    Based on my FLASH4 and FLASH030 utilities
+    GPL 2 License
+*/
+
+#include <errno.h>
+#include <fcntl.h>
+#include <stdbool.h>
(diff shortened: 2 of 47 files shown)


EtchedPixels:
Two things of this lot I think need tweaking the rest looks very good to me, especially the bootdev bits

1. The assumption Zeta v2 will always be below 48K I think is iffy should someone enable networking on it. At that point it might push the 48K limit to judge from Z80pack and COCO3

2. The /dev/physmem should be /dev/mem (and the existing /dev/mem should be /dev/kmem). It also shouldn't be dependent on the ram disc so probably wants to be in devsys.c with a platform/arch or banker provided readphys/writephys or somesuch ?

willsowerbutts:
Alan

Thanks for looking at these patches.

1. Zeta2 48KB image limit: There were two motivations for imposing the 48KB limit;

a) difficult to load kernel images larger than 48KB. Kernel needs copying to specific pages but there is no way to read back the MMU config on the Zeta2. The earlier code did a brute force search of physical memory to find the pages mapped in -- but it assumed that the kernel was loaded in two 32KB chunks each occupying two adjacent 16KB pages. This held true for ROM booting and for the cpm-loader, but with the new fuzix-loader mechanism this assumption fails. We always have to unmap at least one page in order to free up a bank to examine or copy to other pagee, so the alternative plan I considered was to go searching for one (or more) of the 16KB pages -- hash the contents of one or all pages and then search physical memory for pages matching those hashes. A bit messy and hard to entirely preclude a hash collision. The 48KB limit avoids all this and makes the code very simple since we know the top bank can be freely remapped.

b) Once the kernel is running the top bank is common memory shared with the process, so cannot contain kernel code or data structures except in the top 4KB. So the kernel is practically limited to 48KB low+4KB high anyway.

I would like to remove this 48KB limit in the future but I think it will require something more invasive. I am considering an approach where the kernel resides in multiple banks -- a bank for FS code and buffers, a bank for networking code and buffers, etc. At this point we'd need to rewrite the loader anyway.

I am happy to write an alternative if you think (b) will be solved in the future. Also most happy to entertain your suggestions for how to crack challenge (a).

2. Funnily enough an earlier version of this code did implement this through devsys.c! I rewrote it this way to take advantage of code size from unifying /dev/mem with /dev/rd*. I will have another run at rewriting it that way if you'd like me to -- I have had some new ideas for how to make it efficient.

willsowerbutts:
I've done a draft addressing your second point. I'll test it this evening and open a new pull request with new patches.

willsowerbutts:
I'm closing this pull request as the new "features2" #475 is superior.

PR #479, opened 2017-01-22
Kernel: devfd: Optimisations for reliable operation on 6MHz Z80

willsowerbutts:
The driver would previously not work below about 10MHz.

Tested on zeta-v2 (Z80) at 6MHz, 8MHz, 10MHz, 18.432MHz and 20MHz.

Tested on p112 (Z180) at 16MHz.

This change also increases the maximum number of times a floppy
operation will be retried before we fail it. It also no longer panics on
floppy errors.

### DIFF
--- a/Kernel/dev/devfd.c
+++ b/Kernel/dev/devfd.c
@@ -125,7 +125,7 @@ static int fd_transfer(bool rwflag, uint8_t minor, uint8_t rawflag)
 	if (devfd_error) {
 		kprintf("fd_%s: error %d track %d sector %d\n",
 				rwflag ? "read" : "write", devfd_error, devfd_track, devfd_sector);
-		panic("fd_transfer");
+		goto failout;
 	}
 
 	if (retc) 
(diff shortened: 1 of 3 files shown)


willsowerbutts:
Just added another patch to spin up floppy drive motors only when the drive is used (instead of always spinning up all motors).

EtchedPixels:
There are some ways to speed the data loops up even more I think. I don't have the code handy but if I remember rightly the classic Z80 4MHz loop for this is

; C data port
; HL = data
; B  = 128 for 128 bytes, 0 for 256 bytes and loop repeated inline twice for 512
wait:
in a,(FDC_STATUS)
add a
jr nc,wait
add a
jp p, error
ini
jp nz, wait

which is 58 clocks of the 64 allowed and does 256 bytes. At 4MHZ you don't have time to check a second loop count so the code has to be repeated if doing 512 bytes (or self modified to poke the first byte of loop 2 with a ret for 256)

vandys:
In fact, the Cromemco counterpart to this problem had a magic hardware location which wait stated you, taking the busy-wait loop timing out of the picture.  It was the only way some of their slower cards could pull off the needed timing.

PR #484, opened 2017-01-30
Update trsload80.s to work on real hardware

BartHM:
In order to run Fuzix on a real TRS-80 Model 4 some corrections have to be made to the bootloader:
- The correct mode was not selected, Video Page #1 needs to be selected. The emulator SDLTRS does not behave correctly; on a real Model 4 Video Page #0 starts on FC00 with a distorted screen as result. So: OUT (0x84),#0x86 is working correctly.
- Stackpointer out of the way on 0xEE00, doesn't overwrite code.
- The ctc6845 registers are not loaded correctly, they must be read in reverse order. So INC HL must be DEC HL and the label must be put at the end of the table.
- The Clear Screen routine was not implemented correctly.
- On a real Model 4 a formatted floppy must be used to boot. This will be a SDDD disk, 40 tracks, formatted for TRS-DOS 6 or LS-DOS 6. These formatted disks have 18 sectors per track, starting on sector 0 and ending on 17. So an extra DEC A is used and a test against sector 18 instead of 19.

With these changes the bootloader is still 256 bytes and now works on SDLTRS and on a real Model 4 128K. A real or virtual formatted floppy is necessary, so the instructions need an update as well.

### DIFF
--- a/Kernel/platform-trs80/trs80load.s
+++ b/Kernel/platform-trs80/trs80load.s
@@ -2,10 +2,11 @@
 ;
 ;	TRS80 bootblock (256 bytes)
 ;
-;	    .org #0x0
+.area	    BOOT	(ABS)
+.org	    0x000
 start:
-	    ld a, #0x06			; kernel map, 80 column, no remap
-	    out (0x84), a
+	    ld a, #0x86			; kernel map, 80 column, no remap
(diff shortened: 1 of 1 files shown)


EtchedPixels:
Many thanks

PR #488, opened 2017-02-12
coconic: change to static address on init, enable 16 bit xfers

beretta42:
make coconic faster, and move i/o base to an multi-pak friendly address.

### DIFF
--- a/Applications/netd/coconic.c
+++ b/Applications/netd/coconic.c
@@ -15,16 +15,17 @@
 #include "uip.h"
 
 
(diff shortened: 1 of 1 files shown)


EtchedPixels:
If you are setting the low bytes of the mac then it ought to be random not fixed - two boxes with the same mac makes a nasty mess.

beretta42:
Actually netd reads and sets the MAC from /etc/netrc.  This patch changes
to I/O base address of the card in memory, making it easier to use with our
multi-pak  cart multiplexer.

Brett

PR #503, opened 2017-05-25
exec(): take PROGLOAD from the binary header

beretta42:
This experiment seems to let me load and run binaries based off the progload address found in the binary header, rather than a static address.  This allow me to load and run bins from stock dragon/coco2 root images. I wanted to redefine PROGLOAD, but unfortunately, "start.c" uses this #define to a great extent.  It seems a redefined macro for PROGLOAD would have to be pretty smart for both usages.

### DIFF
--- a/Kernel/syscall_exec16.c
+++ b/Kernel/syscall_exec16.c
@@ -81,8 +81,8 @@ static int header_ok(uint8_t *pp)
 	p += 3;
 	if (*p++ != 'F' || *p++ != 'Z' || *p++ != 'X' || *p++ != '1')
 		return 0;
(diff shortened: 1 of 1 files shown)


EtchedPixels:
This breaks self-relocating binaries (not that we have any yet except for tests). I think you need to spot the 0 case and set it to PROGLOAD. Otherwise this looks sensible.

beretta42:
Ok.  I should have expected that.  So load page of 0 means "pick whatever".  And pass that value to the binary, for self modifiers.  Interesting,  I remember seeing C64 code that does such things.

EtchedPixels:
thanks

PR #505, opened 2017-05-30
fix msx2 port

retrodeluxe:
With this the msx2 kernel is able to boot rootfs-z80-32 from fuzix.org, 
some random hangs are also gone.

### DIFF
--- a/Kernel/platform-msx2/config.h
+++ b/Kernel/platform-msx2/config.h
@@ -45,6 +45,7 @@
 
 #define CONFIG_SD
 #define SD_DRIVE_COUNT 1
+#define BOOTDEVICE 0x0100
 
 #define MAX_BLKDEV 1      /* Single SD drive */
 #define CONFIG_RTC
--- a/Kernel/platform-msx2/devmegasd.c
+++ b/Kernel/platform-msx2/devmegasd.c
@@ -128,117 +128,58 @@ uint8_t sd_spi_receive_byte(void)
  * Target page is always mapped to slot_page2, and the target address offset accordingly.
  *
  */
-bool sd_spi_receive_sector(void) __naked
+void sd_spi_txrx_sector(bool is_read)
 {
-    __asm
+    uint16_t addr, len, len2;
+    uint8_t *page;
+    uint8_t page_offset;
 
-    ; map sd interface
-    ;
-    ld a,(_slotmfr)
-    call _mapslot_bank1
-    ld a, #MSD_PAGE
-    ld (MFR_BANKSEL0),a
+    addr = ((uint16_t) blk_op.addr) % 0x4000 + 0x8000;
+    page_offset = (((uint16_t)blk_op.addr) / 0x4000);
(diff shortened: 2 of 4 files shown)


EtchedPixels:
I am not sure it's always safe to just di rather than save the IFF state as you will end up leaving interrupts disabled when they should not be. 

The SD card one wants a different fix. The underlying problem is that it's trying to use the page number to work out where to read whereas the kernel passes a request for raw I/O - which expects the user map to be used.

Basically it should be doing

IS_USER_OFFSET = 0: use kernel mapping
IS_USER_OFFSET = 1; map the user process, do xfer, map the kernel 
IS_USER_OFFSET = 2; map the page specified, do xfer, map the kernel (and will be aligned)

That also means sd_spi_receive/transmit need to be in common memory.

Alan

retrodeluxe:
> I am not sure it's always safe to just di rather than save the IFF state as you will end up leaving interrupts disabled when they should not be.

right, I guess I can use __hard_di and restore.

> The SD card one wants a different fix. The underlying problem is that it's trying to use the page number to work out where to read whereas the kernel passes a request for raw I/O - which expects the user map to be used.
> 
> Basically it should be doing
> 
> IS_USER_OFFSET = 0: use kernel mapping
> IS_USER_OFFSET = 1; map the user process, do xfer, map the kernel
> IS_USER_OFFSET = 2; map the page specified, do xfer, map the kernel (and will be aligned)

> That also means sd_spi_receive/transmit need to be in common memory.

the sdcard interface spi ports need to be mapped at a fixed address in page1 before doing the transfer, which means that I need to set that page before doing the transfer and can only map the process (or the kernel) in pages0, 2 and 3.

So even if sd_spi_receive/transmit are in common, still have the same problem unless I use an intermediate buffer also in common and do a two step memcpy.

I think I can just work out the logic to handle blocks that fall between pages.

EtchedPixels:
For map_save/map_restore you know we are in di state. It might be worth having a map_kernel_di or similar as the hottest paths are already IRQ off and that would avoid the save/restore overhead in many cases.

I keep forgetting how demented I/O is on MSX.

In which case yes for the IS_USER_OFFSET 1 case you need to actually handle page crossings in a request and that while the length will be a multiple of 512 the address can be arbitrary. Which reminds me I need to fix this on the NC100/NC200 as well 8)

retrodeluxe:
Hi,
 I have found that the actual problem with map_process and the interrupt is that the update of map_table is not atomic: If an interrupt happens between one of the page switches and the update of the map_table entry; then map_save takes the wrong values to restore the pages after interrupt; and on return, even though map_process corrects the map_table, it keeps the wrong page in the mapper register.
 
  so if we just save map_table at the beginning of map_process then, in case of interrupt, map_restore takes care of setting the pages properly no matter when map_process is interrupted. 
This seems to work and requires changes only in msx2.s

EtchedPixels:
Merged - thanks

PR #529, opened 2017-10-22
Kernel: quick pointer fix.

beretta42:
Quick somebody hand me a '&' !  :)


beretta42:
keep screwing up my commit branches :(

PR #534, opened 2017-10-24
bank16k: report correct pagemaps used

beretta42:
pfptr goes up as free maps are added to the free list.  return the compliment instead.


beretta42:
ick... needed more testing.  512k coco is now misreporting memory usage...

EtchedPixels:
Good point on the fsck.. will fix that right now!

PR #561, opened 2017-11-17
Uip bind

beretta42:
hook up bind() event, and modified uip to optionally choose a local port.


beretta42:
I must do more work.  Applying this patch wonks TCP. :(

PR #569, opened 2017-11-25
Quick fixes for netd

beretta42:
"user_request" flag does nothing in the UDP handling, and UDP doesn't need any complicated closing anyway, so release and free up UDP struct ASAP after kernel's close().


EtchedPixels:
strlcpy is better than strncpy btw unless you need to zero all the rest of the buffer

Issue #555, opened 2017-11-13
netd mishandles TCP reset, some types of close and bind then connect


EtchedPixels:
A reset should close the socket for read and write

A close should break the connection between the uip 'socket' and the kernel one and close the kernel one immediately. If data is pending for input then tcp reset the link, if not then let it cycle through close states and reset if any more data arrives

uIP needs tweaking so that it uses the bind() provided local socket if presetn when we connect().

For performance reasons we could do with teaching uIP to be able to answer the question "how many timer cycles until my next wakeup"

And we need to move to the most recent uIP code base (in Contiki)

beretta42:
re: close to immediately move kernel socket to CLOSED:  net_native.c looks to be the place to make this change.  Close the socket, but leave the SD struct around for message buffering.

beretta42:
Re bind: Lemme check my passive open branch.  I sorta remember implementing bind() there.   I think its required for passive opens, no?

EtchedPixels:
The close needs to go via net_native because net_native needs to know when it closed, also when the Wiznet offloaded tcp/ip is done I'm fairly sure we need to block (momentarily only) on doing the close.

So really the 'I'm closed' 'ok you're closed' handshake does need to go through netd

EtchedPixels:
I think I have the reset/close side nailed down now.

EtchedPixels:
The bind logic is handled in the kernel but we need to tweak uIP to pass the local address from the kernel/netd into uip_connect() instead of UIP making it up.

beretta42:
I commensed tweaking uip for bind support.  I didnt have time to Test/wireshark the changes but soon.

beretta42:
I think a bind on a udp sock should move the sock into connected state?

EtchedPixels:
As uIP sees it quite possibly.

In the Unix world INADDR_ANY, port bind() means 'receive anything to this port and addr/port means 'receive anything to this port/address pair'

Connect() in Unixland means 'send to this address/port only'

And I'd need to look up whether it means received frames must have that address. I don't think so but need to check!

beretta42:
Hmm.. i dont think "sendto" actually sends any signal to netd to change uip's remote address/port.  It will send a "write" event, then i guess i can make netd apply the remote addr every write.  Ick

beretta42:
[...]

PR #571, opened 2017-12-07
netd: let Fuzix choose local port

beretta42:
as per issue #555
"uIP needs tweaking so that it uses the bind() provided local socket if presetn when we connect()."

if we pass a local port of 0 to uip, it'll choose the local port, else it'll use fuzix's idea of the port no.  is there a reason even to bother with uip picking it's local port?


EtchedPixels:
There should no case where 0 ever gets passed.

PR #572, opened 2017-12-10
NC200 floppy boot support

davidgiven:
This adds a platform-nc200 which contains a fairly crude proof-of-concept that will generate a bootable floppy disk containing an NC200 kernel. It only gets as far as displaying the banner, and sometimes not even then, but it's enough to demonstrate that it'll actually load and run.

This also converts the nc100 bootblock.s to sdasz80 syntax and adds a makefile rule for it. (And adds the few instructions needed for the floppy boot to work.)

I'm not particularly happy about the way the new platform uses the nc100 files, but right now I think it's more important to disturb the nc100 stuff as little as possible.


EtchedPixels:
Agreed - but we can fuse NC100/NC200 if appropriate once it works

PR #575, opened 2017-12-18
coco3: take cmd line for default video mode

beretta42:
Because life with a NTSC T.V. at 80 columns sucks.


EtchedPixels:
Can we have the cmd line options in lower case please - otherwise looks good. Can you also switch with the mode ioctl/cmd ?

beretta42:
I choose uppercase as its marginally easier to pass from BASIC.  Yeah, the mode ioctl works fine for me.  I guess strcasecmp would work.

PR #584, opened 2018-02-03
Adjusted and updated Documentation directory!

marioaugustorama:
Only updates into documentation directory.


EtchedPixels:
I don't think having a TODO list like that tracked in git is a good idea because it should be auto-generated if we have it at all. If the Makefile generated it then that would be great.

marioaugustorama:
I even have the makefile written for this purpose, but no commit yet!

PR #587, opened 2018-02-20
build: add fsck -y passing to buildscript

beretta42:
Allows one to use Neal's filesystem builder non-interactively.


EtchedPixels:
Yep Tormod suggested this too but I didn't want to merge it until the ucp bugs were fixed


Now they are so in it goes

PR #589, opened 2018-02-24
coco3: implement sd card.

beretta42:
Implements the low level bits needed to use Kernel/dev/devsd on the CoCoFPGA's sd card.


EtchedPixels:
Please add a patch to document the reason for the "nop" though

PR #590, opened 2018-02-25
coco3: no star trek until we have 6809 math, or fixed point startrek

beretta42:
Not yet :(

### DIFF
--- a/Applications/games/Makefile.6809
+++ b/Applications/games/Makefile.6809
@@ -19,7 +19,7 @@ SRCSNS = qrun.c fortune.c
 SRCS  = adv01.c adv02.c adv03.c adv04.c adv05.c adv06.c adv07.c \
         adv08.c adv09.c adv10.c adv11.c adv12.c adv13.c adv14a.c adv14b.c \
         myst01.c myst02.c myst03.c myst04.c myst05.c myst06.c myst07.c \
-        myst08.c myst09.c myst10.c myst11.c fortune-gen.c startrek.c
+        myst08.c myst09.c myst10.c myst11.c fortune-gen.c
 
 OBJS = $(SRCS:.c=.o)
 OBJSNS = $(SRCSNS:.c=.o)


EtchedPixels:
They should be in your gcc support library. Look for __addsf3 etc.

beretta42:
Ah.. I never looked at gcc floating point stuff.  There is nothing in our libgcc.a about floats... anyway I made fuzix make libm.a, but the very first file, acos(), dies complaining about not having uint16.

EtchedPixels:
You need to start by building gcc/libgcc/soft-fp for 6809.

EtchedPixels:
Tweaked - libm now sort of builds, but there are some things we need to take care of.

EtchedPixels:
What may be needed is to redo the double section so that for 6809 it's 

union {
    double value;
    uint32_t bits[2];
};

and change the macros (actually some get simpler!). Also none of our stuff uses INSERT_WORD64 so it can go away.

EtchedPixels:
Ok I've pushed what ought to work as the full higher level maths lib if combined with the gcc soft float library. Not tested it other than compiling.

beretta42:
I might have softfloats.  Libgcc now seems to have floats.  printf now complains about lack of _fnum.  Where he supposed to live?

beretta42:
I turned off -DBUILD_LIBM, got a bit further.  printf stopped complaining. Now startrek.c is failing on sqrt()...The macro seem to suggest that should have been replaced with sqrtf().  Math.h: #ifdef double.  Seems like gcc is not including the double to float headers.

beretta42:
Bah.. i see double is being refined in stdint for us 6809'ers... my local git much be a few commits off.  Cross fingers.

beretta42:
Ah.. ok. so i have to compile gcc with doubles... missed that in my haste also.  I've grepped sdcc,gcc, and fuzix for vprintf's helper "fnum()"  without luck.  Where's he hiding?

beretta42:
While attempting to build libgcc with doubles:

./../../gcc-4.6.4/libgcc/../gcc -I../../../../gcc-4.6.4/libgcc/../include  -DHAV
E_CC_TLS -DUSE_EMUTLS -o _pack_sf.o -MT _pack_sf.o -MD -MP -MF _pack_sf.dep -DFI
NE_GRAINED_LIBRARIES -DL_pack_sf -c _pack_sf.c  -save-temps
In file included from _pack_sf.c:47:0:
../../../../gcc-4.6.4/libgcc/../gcc/config/fp-bit.h:101:1: error: unable to emulate 'DI'
In file included from _pack_sf.c:47:0:
../../../../gcc-4.6.4/libgcc/../gcc/config/fp-bit.h:114:1: error: unable to emulate 'DI'

DI is 64 bit integer support?   :(

EtchedPixels:
Sounds like it's getting closer - I got stuckj away due to weather so I'll take a look at this once I'm back in order

beretta42:
Thanks Alan, 6809 builds again.  'startrek' compiles, and mostly works, except the actual floaty parts.

EtchedPixels:
Yeah.. the two 32bit halves are probably backwards or something. Will need some debug with simple float apps

PR #607, opened 2018-04-21
Rearrange platform-nc100 and platform-nc200

davidgiven:
This combines the platform-nc100 and -nc200 directories into a single common code directory (platform-amstradnc) with subdirectories for the platform-specific files (currently, build scripts only). This means that the two targets are now TARGET=amstradnc/nc100 and TARGET=amstradnc/nc200, which is a bit clunky, but this seems better than having three different top-level platform directories.

I've tested the NC100 platform on nc100em and it at least boots to the prompt (I haven't tried it with a root filesystem). The NC200 floppy boots but dies on startup (the rest of my port needs merging across).


EtchedPixels:
Actually I think hierarchically naming some of them might make a lot of sense (eg Z80pack too)

PR #608, opened 2018-04-23
NC200 screen and floppy support

davidgiven:
This adds support for the NC200 screen, my floppy driver, and updates the documentation.

It all seems reasonably solid, although h doesn't work; it hangs when it tries to run programs. (With interrupts on.) Everything else appears to be fine, which makes me wonder whether this is my fault or not.

The floppy driver is painfully slow --- even without the block device framework, and with `CONFIG_LARGE_IO_DIRECT` set, I'm never receiving multisector or raw requests; binaries are all loaded one sector at a time via a kernel disk buffer...

The biggest problem is that the 720kB floppy isn't really big enough for a decent root filesystem!


EtchedPixels:
Ok the direct I/O should at least now happen

PR #612, opened 2018-04-30
coco3: first stab at sd driver for Roger Taylor's coco on a nano

beretta42:


EtchedPixels:
As well as loop unrolling you might want to look at the timing patterns. For some systems if you know the CPU clock rate and the SD clock rate it's cheaper to just nop than check the status. And in fact in some cases the SD is so much faster than the CPU you know it won't be busy 8)

PR #628, opened 2018-06-16
util: add xxd from the VIM project

beretta42:
Works out of box from the VIM project.  Could probably use a haircut.


EtchedPixels:
what useful can this do that od can't ?

beretta42:
Uhm.. i like that it has "-r".. and will reverse the process.  Like a hex editor for "free".

beretta42:
ok... I did near 0% effort on this one. :)

PR #630, opened 2018-06-26
Z80 fuzix ccopts

piso77:
One of the major slowdown in the z80 build is the O2 optimization level.

This pull request introduce a new global $FUZIX_CCOPTS variable that encapsulate the cc optimization level (and possibly other cc options), and is reused by all the z80 Makefiles down the road.
By default this pull request doesn't change the actual O2 level, but can that on the command line:

$ FUZIX_CCOPTS=-O0 make ...

On a sluggish Intel Atom c750 4 cores / 8 threads, make -j4 goes down from ~2hours to ~30 minutes.
On the Github / Google Cloud default instance for Travis CI (2 x 2.6Ghz Xeon with 7.5Gb of ram) i could rebuild the entire z80pack in less than 15 minutes, without any swap usage.

While not perfect (this pull only cover the z80 side of the build), it is mandatory for another feature that i would like to (re)enable: the Github / Travis continuos integration - that has a time limit of 50 minutes per job for the free instance, and without this hack we constantly blow that time limit.

### DIFF
--- a/Applications/MWC/cmd/Makefile.z80
+++ b/Applications/MWC/cmd/Makefile.z80
@@ -1,5 +1,5 @@
 FCC = ../../../Library/tools/fcc -m$(USERCPU) $(Z80_PLATFORM)
-FCCOPTS = -O2
+FCCOPTS = $(FUZIX_CCOPTS)
 
 .SUFFIXES: .c .rel .y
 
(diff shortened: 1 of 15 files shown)


EtchedPixels:
I do need to look at tuning the values as well - sdcc has gotten much better at generating decent code with lower optimisations (except for longs which are in a bit of a sad state). The latest sdcc snapshots also fixed the crazy memory usage of things like uip.c.

This definitely makes sense for Z80. For the other platforms I don't think it's relevant - gcc and cc65 are way faster.

PR #638, opened 2018-07-10
Resurrect the m68k arch, and hook v68 to Travis

piso77:


EtchedPixels:
Thanks - the 68K port is very much a prototype. I'm actually surprised the normal build setup works with so few patches !

piso77:
Well, it builds, not sure it does anything useful... :)

But one thing: when packaging the m68k-uclinux toolchain, i found that elf2flt generates its own ld variant (m68k-uclinux-ld.elf2flt), but i didn't use it and kept the one generated from binutils (m68k-uclinux-ld), not sure if that impacts the build in any way...

PR #679, opened 2018-12-09
rc2014: fixed missing functions when ppide enabled

c32ff0:
ppide_rbc.c has #ifdef CONFIG_PPIDE so we don't add useless code if PPIDE isn't configured


EtchedPixels:
Thanks - applied

PR #692, opened 2019-03-09
Use byacc not yacc to generate /bin/find grammar

nickd4:
Bison output won't compile, at least on the Z80 port. Even if it worked, I think it is risky trying to use bison output in any kind of small context since it's pretty bloated.

I used byacc since it's readily available on a ubuntu system. Ideally we'd use something like V7 yacc, but this was just a quick solution to get going again for the time being.

This patch needs to be tested for correctness, I will get to it eventually but for the time being I just want to be able to complete the compilation process without errors.

### DIFF
--- a/Applications/MWC/cmd/Makefile.z80
+++ b/Applications/MWC/cmd/Makefile.z80
@@ -36,6 +36,10 @@ expr.c: expr.y
 
 find.c: find.y
 
+# Nick
+.y.c:
+	byacc -o $@ $<
+
 .c.rel:
 	$(FCC) $(PLATFORM) $(FCCOPTS) -c $<
 


EtchedPixels:
If you call "byacc" it then breaks everywhere else. This one is actually covered in the INSTALL document at the top level.

Fuzix expects yacc to be yacc not Bison. if your system for some reason thinks
that executing yacc should run bison you'll have a build failure. For Debian
this can be fixed with

 - sudo apt-get install -y byacc
 - sudo update-alternatives --set yacc /usr/bin/byacc

I did look at the V7 yacc and I agree something like that might be useful or maybe the MWC one from Coherent.

nickd4:
Okay. I didn't know how to do the update-alternatives voodoo, though I have seen it around. I will check the INSTALL document more carefully, though I've found workarounds for the minor issues I encountered.

PR #694, opened 2019-03-11
usermem: cast so gcc doesn't complain

beretta42:


EtchedPixels:
Already in my tree but correct

Issue #700, opened 2019-04-16
direct I/O & swap Q's


jcw:
(me again) - I'm trying to debug some I/O issues. Two problems I'm running into:

* direct I/O is not working -  it hangs after reading what looks like the last block of `init` in raw=0, right after all the raw=2 calls, which are loaded to the right address and bank

I'm running without direct I/O for now. The second puzzle, which I think is more urgent, is:

* swapouts seem to be ok - 120 blocks are written to the swap device, but swapin fails:

```
hd0: W @ 1	#1 addr 0:AC46 desc 0xC28F raw 0
Swapping in B4F8 (page 4), utab.ptab B541
hd1: r @ 7320	#120 addr 4:0000 desc 0xC28F raw 2
Swapped in B4F8 (page 4), udata.ptab B541
0049_switchin: FAIL
```

What I see is that it gets swapped in to a different bank, but that should be fine IMO.

The way I'm running the swap test is to run nested shells until I see a swapout, then exit back from each shell until a swapin is required. That immediately runs into the above issue.

I'll read through some more of the code and docs, but if anyone has tips on how to chase this ...

PS. This uses z80pack mapping: 60K banks (1 kernel, 7 user):
: SWAPTOP: 0xF000, PROGTOP: 0xEE00, COMMONMEM: 0xF400.

EtchedPixels:
Where is your udata stash located ?

I would also dump the relevant bits of the disk image after the swap and see if what you have is valid. There is a lot (almost everything) in common between the direct I/O and swap I/O path in banked mode so they may be connected.

The switchin error occurs when you swap in a process or restore its data and the udata.ptab is not the right one. That often means your swapin reads actually went nowhere or the wrong place - which would also fit your direct I/O fail

jcw:
Ok - U_DATA_STASH = 0xEE00 in kernel.def, same as PROGTOP in config.h - MA_SIZE is 0xF000.

I'll check that the process data itself is actuall y going to the swap disk and being read back in. That would indeed explain what's going in.

jcw:
Found it. Was passing some data into to the disk driver via a struct that got "banked away". Solved by placing the info on the stack instead. Doh (in hindsight).

And indeed, this also fixed the direct I/O bug.

Thanks for pointing me in the right direction. Checking the actual I/O activity did the trick!

EtchedPixels:
Been there, done that with variables a few times

jcw:
[...]

Issue #699, opened 2019-04-14
trying to debug a new z80 port, some Q's


jcw:
Hello - I'm trying to debug a new port to a custom Z80 emulator ([link](https://jeelabs.org/2018/z80-zexall-bp/)) - my code is different from z80pack, in that it supports setting the starting address of the common bank (set to 0xF000 on init), as well as setting/getting the bank (currently 8 banks, with bank 0 used for the kernel).

What I'm getting so far is:

```
FUZIX version 0.3pre1
Copyright (c) 1988-2002 by H.F.Bower, D.Braun, S.Nitschke, H.Peraza
Copyright (c) 1997-2001 by Arcady Schekochikhin, Adriano C. R. da Cunha
Copyright (c) 2013-2015 Will Sowerbutts <will@sowerbutts.com>
Copyright (c) 2014-2019 Alan Cox <alan@etchedpixels.co.uk>
Devboot
484kB total RAM, 420kB available to processes (15 processes max)
Enabling interrupts ... ok.
Mounting root fs (root_dev=256, ro): HD wr 0 mem 0xAD43 pos 1
HD wr 0 mem 0xAB3B pos 2
OK
Starting /init
HD wr 0 mem 0xA933 pos 256
HD wr 0 mem 0xA72B pos 19
HD wr 0 mem 0xA523 pos 294
```

Those "HD ..." lines are disk access debug prints.

The difference with z80pack's banked memory, seems to be that I don't really have a way to turn banking off, once enabled, nor write-protect anything - I can merely switch between banks 0..7.

Evidently, the logic in `z80pack.s` needs to be adjusted to this, but I haven't figured out yet how I/O on port 23 should be handled: can I simply comment out those protect/unprotect calls?

Also, what sort of disk I/O should I expect as it's starting `init` ? I see some bank switches between banks 0 and 7, but I assume the next thing that should happen is reads into bank 7?

(FWIW: many things are not yet implemented, such as no interrupts or tty input yet)

EtchedPixels:
I would expect to see a tiny amount of kernel targetted I/O mounting which you have. It then should read the path and inode to find /init and then load the first block of init into a kernel buffer, and the rest directly into user space if you have direct I/O enabled.

You may find the sbcv2 port is much better as a guide. It's a bit different in split (32K/32K) but heavily commented and has no weirdness like memory protection.

[...]

PR #702, opened 2019-04-19
new JeeRetro build

jcw:
Please consider adding this "platform-jeeretro" subdirectory to the FUZIX project. It's meant for use with the Z80 emulator at <https://git.jeelabs.org/jcw/retro/src/branch/master/black407-z80ex>.

See issues #699 and #700 for relevant discussions and fixes that led to this port.

There's still an issue when running on an STM32F407 µC, but the port builds and works in MacOS and Linux. It's now set to use only 2 user banks so it will also fit in 184 KB RAM on the F407.


EtchedPixels:
Can you clean up the README to reference your platform and perhaps a URL link to the I/O port model it uses. Maybe grep the code for z80pack and fix the other references as appropriate.

Also I can't help thinking it would be nicer if the max ram was not hard coded but an in() reported max banks ?

Otherwise looks good - would be nice to know why are doing the stuff you are with the tty read (in and write to tmpReg), if you are getting a miscompile there with sdcc 3.9.0-rc that's a nasty and one that ought to go to Philipp to get nailed before release.

jcw:
Ok, will do. I don't like the I/O model for system calls, but inherited it from the z80emu example. Have been considering switching to "HALT + .db XYZ", but that will run away the moment an app uses HALT. The cleanest way would perhaps be to "virtualise" all I/O and HALT, and maybe also DI/EI/RETI/RETN.

Yes, hard-coded banking is messy, and may cause problems later. The tty read stuff is just me hacking my way through to get things going. I'm pretty sure I can clean it up a bit, now that it's working.

BTW, the reason I'm doing this is because the z80emu project was a nice / small / clean starting point, which works nicely on a range of µC's. The F407 has 192K ram, hence this attempt to port FUZIX to it. There's overlap with the RC2014 emulator, evidently.

jcw:
Re tty read w/ tmpReg: ah, yes, that's because I don't know the SDCC conventions, i.e. how to pass args in and out of an __asm block. I didn't want to move these routines to a .s file - or should I?

EtchedPixels:
SDCC does this in a kind of weird 8051ish way

__sfr __at 0x01 foo;

then x = foo generates an in

It would have been far better if it was pointer based as you can't do maths on it nicely.

The lowlevel-z80 runtime also provides in() and out() as function calls for the other cases.

EtchedPixels:
Merged.

I will push a few other trivial things (several of which I need to also push for z80pack!)

PR #713, opened 2019-05-21
coco3: stack blast on mmu bank copies

beretta42:
This stack blasting copier is 3.8 cycles per byte copied. The old code was 7.3 cycles per byte. This method could also be helpful in the dragon-mooh port @tormodvolden.


EtchedPixels:
Doesn't this break support for using the fast interrupt stuff or is the code such that if it pushes/pops onto the working copy stack it'll be fine (even if you have to make the last 32 bytes or so the old way) ?

beretta42:
oops... the fast interrupts didn't occur to me.  fixed as suggested.  I got lucky and use S as my destination pointer. revision pushed.

PR #731, opened 2019-07-05
vile: fix read

beretta42:
the old code still bombs out if loading a small file, as even after loading, "size" is still larger than int_max and doread() fails with EINVAL.


EtchedPixels:
thanks

PR #733, opened 2019-07-06
yaz180 - fixes - no merge

feilipu:
a set of fixes for yaz180:

- ensure ram is mapped in contiguous banks
- match config to physical RAM
- a bit of tidying on the way through


Kernel/cpu-z80/rules.mk
```diff
@@ -1,9 +1,9 @@
 export CROSS_AS=sdasz80
 export CROSS_LD=tools/bankld/sdldz80
 export CROSS_CC=sdcc
-#export CROSS_CCOPTS=-c --std-sdcc99 --no-std-crt0 -m$(CPU) -I$(ROOT_DIR)/cpu-$(CPU) -I$(ROOT_DIR)/platform-$(TARGET) -I$(ROOT_DIR)/include --max-allocs-per-node 1000000 --opt-code-size --stack-auto --constseg CONST
-#export CROSS_CCOPTS=-c --std-sdcc99 --no-std-crt0 -m$(CPU) -I$(ROOT_DIR)/cpu-$(CPU) -I$(ROOT_DIR)/platform-$(TARGET) -I$(ROOT_DIR)/include --max-allocs-per-node 200000 --opt-code-size --stack-auto --constseg CONST
-export CROSS_CCOPTS=-c --std-sdcc99 --no-std-crt0 -m$(CPU) -I$(ROOT_DIR)/cpu-$(CPU) -I$(ROOT_DIR)/platform-$(TARGET) -I$(ROOT_DIR)/include --max-allocs-per-node 30000 --opt-code-size --stack-auto --constseg CONST --peep-file $(ROOT_DIR)/cpu-z80/switch.peep #--fverbose-asm
+#export CROSS_CCOPTS=-c --std-sdcc99 --no-std-crt0 -m$(CPU) -I$(ROOT_DIR)/cpu-$(CPU) -I$(ROOT_DIR)/platform-$(TARGET) -I$(ROOT_DIR)/include --max-allocs-per-node1000000 --allow-unsafe-read --opt-code-size --stack-auto --constseg CONST
```
Jo-Con-El: When adding the `--allow-unsafe-read` flag here and the next two lines, you inadvertently removed the space in the `--max-allocs-per-node` parameter.

feilipu: Thanks. :+1: 

This pull needs more work to be completed, and I've been distracted, but the space in `--max-allocs-per-node 30000` sometimes causes problems.

It works better to put the parameter directly following the config flag `--max-allocs-per-node30000`, otherwise the `30000` gets parsed as an incorrect config flag.

It looks ugly, but that's life. :man_shrugging:

PR #734, opened 2019-07-06
implement platform and userspace side of /dev/sys

beretta42:


EtchedPixels:
Can we keep to the normal pattern of ioctl names a bit more - and also use the names in the coco3 code.

How does

CPUIOC_6809SWI2
CPUIOC_6809SWI3

sound ?

beretta42:
yeah.. you're stupid to let me name any macro/define/variable in mainline code... I'm horribly bad at that sort of thing.  Where should the kernel's #defines go?  I didn't see a place that was obvious, or a .h file in Kernel/include that made sense.

EtchedPixels:
Just create a kernel/include/cpu_ioctl.h or somesuch and then it needs to get propagated to be Library/include/sys/cpu_ioctl.h for userspace to pick up.

EtchedPixels:
Thanks

PR #742, opened 2019-08-22
tty: turn on IGNCR support in kernel

beretta42:
This looks to be supported. I don't see anymore that are left out, on the input flags anyway.


EtchedPixels:
Thanks - both applied

PR #770, opened 2020-02-11
Header fix

beretta42:
some easy fixes for the new bin header changes.  I get to init now, but now my sh is complaining of "membad", or just crashing.


EtchedPixels:
Applied - I missed the 6809 segments and the fact it could get spaced further.

membad suggests to me that the header isn't getting filled right so the BSS and allocations get mucked up and the sh memory allocator craps itself

PR #776, opened 2020-03-23
msx2: add missing _udata

retrodeluxe:
this appears to fix msx2 platform down to v0.3.1, it remains broken in
 more recent versions.


EtchedPixels:
Thanks

PR #791, opened 2020-06-04
sc126: add new SPI defines for z80softspi

mennozon:
These were missing for sc126, copied from commit f7d0ed


EtchedPixels:
SC126 is different as it does SPI via the Z180 CSIO. I'm working on other Fuzix and retrocomputing things at the moment but I'll deal with this when I do a later pass before a 0.4 release.

mennozon:
Thanks, it fixed the compile error but I was not aware about the difference.

I'll drop the pull request and read up more about the hardware!

Issue #158, opened 2015-03-30
User space does not build (nc100 port)


rhaberkorn:
First of all, thanks for all your efforts!
I'm trying to prepare a nc100 card image for playing around with it in nc100em. I will also try it on my real NC100 once I get my hands on one of these expensive PCMCIA memory cards. I'm relatively new to z80 programming and the FUZIX build process is very poorly documented, so please excuse any stupid mistakes on my side.

I am running Ubuntu 10.04 with a hand-built SDCC:

    SDCC : z80/z180/gbz80 3.4.3 #9201 (Mar 29 2015) (Linux)

So far I think I've managed to build the kernel "image" (fuzix.bin). First a few remarks on the nc100 kernel build process:
* `nc100.def` is missing. Given the commit log, I assume it must contain something like `VIDEO_BASE .equ 0x7000`, `VIDEO_BASEH .equ 0x7000` (what's that for, anyway?) and `VIDEO_SIZE .equ 0x1000`. You should also add the corresponding `nc200.def`.
* the `Kernel/platform-nc100/Makefile` image-target refers to `bootblock.cim`. This looks like it must be generated from `bootblock.s` with `zmac`. Any reason for not adding this step as a Makefile rule?
* Also I think that the image rule should be modified/fixed and read `dd if=bootblock.cim of=../fuzix.bin bs=1 seek=256 conv=notrunc`

If I understand correctly, the `fuzix.bin` is already a PCMCIA card image, albeit incomplete. Something like a rootfs should be added. I see that this can be done with the `mkfs` and `ucp` tools from the `Standalone` subdirectory. But in reality I assume that the `Standalone/filesystem-src/build-filesystem` script should be used instead of doing it manually. This needs all of the user space tools of course. Unfortunately, the `Applications/util/Makefile` won't just work. Right, we need a toolchain first! It appears to be hidden in `Library/`. Just making it however does not appease the `Applications/util` build proces as it's referring to files in `/opt/fcc/`. So I followed the not-so-outdated instructions from the [wiki](https://github.com/EtchedPixels/FUZIX/wiki/Building-User-Space). I think it should be renamed to "Building the toolchain". But now I get the following build errors from `Applications/util/Makefile`:

```
$ make 
../../Library/tools/fcc -O2  -c banner.c
../../Library/tools/fcc -O2  banner.rel -o banner

?ASlink-Warning-Undefined Global '_strcpy' referenced by module 'vfprintf'

[...]

PR #798, opened 2020-07-25
Make LIBZ80 autodetection error check

oholiab:
Helps bubble up one of the errors from #158 

I was experiencing the same problem (that the library was not being correctly detected) on Arch Linux because the actual binary is installed at `/sbin/sdcc` and `findsdcc` assumes that the libs can be found from a relative path.

Rather than trying to include packaging specific logic to `findsdcc`, I've added a simple check in `libclean` to hard fail if `LIBZ80` is not populated with a valid directory path and to instruct the user to manually set the variable to the correct path.

I've tested this code change by running `make` on my Arch install and having it fail, and then manually setting `LIBZ80` (before the check) and running again and it's now passing the check and continuing the build.

### DIFF
--- a/Library/tools/libclean
+++ b/Library/tools/libclean
@@ -6,6 +6,11 @@
 #
 CPU=$1
 LIBZ80=$(../../Kernel/tools/findsdcc $CPU)
+if [ ! -d $LIBZ80 ]; then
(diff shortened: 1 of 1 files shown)


EtchedPixels:
Sorry this one took so long

PR #803, opened 2020-12-01
[WIP] Port to the TI-84+ Silver Edition graphing calculator

ddevault:
Opening this up a bit early to establish a line of dialogue on some design points.

The initial port should be able to get away with having a read-only filesystem from Flash, but a read/write driver would naturally be good to see in the future. There's only 2MiB of storage, so a conservative filesystem image will be necessary.

One problem I've been considering is swap. The current memory map will only allow for 2 processes to run at once from RAM. We could get more with two approaches: attempt to discern how many pages a process is actually using, and allocate sub-pages as appropriate; or swap to Flash. The latter would be pretty slow, but more importantly, the Flash chip is only rated for 1M write/erase cycles. How much does swap thrash the storage device on FUZIX?

Other points of interest with this port would be trying to hack together an RS232 serial port out of the 2.5mm jack bit-bang interface, and doing something interesting with the USB OTG port. It might be possible to put the filesystem on USB mass storage, for instance, or set up a USB-to-ethernet adapter once the network stack matures.


EtchedPixels:
You don't want to swap to flash, you really don't 8) It won't usually thrash it but there are times it will. Sounds like you'd need to use the USB OTG interface for swap/file system. Two ;processes in RAM at once seems sufficient even with a fairly slow swap device - for single user anyway.

The tty interface for a non-blocking pure graphical console is btw pretty simple. Your kputchar is your tty output char routine for the console port, and it's always ready for output, always has carrier and has no configurable serial properties so the setup is empty too.

The keyboard driver then just needs to queue received chars.

Looks a good starting point.

ddevault:
Yeah, the tty interface is going to be pretty straightforward. There's also an external keyboard device available for this platform which would benefit from a driver. The main frustration with the tty is just the lack of real-estate on the screen, it's only 96x64.

davidgiven:
FWIW, as part of the ESP8266 port there's now a version of the Dhara FTL library in the source. I've only used it on 32-bit systems but if it works on the Z80 it will let you put the filesystem on the flash with proper wear levelling. It does appear to require about 30% of the storage as overhead. There's also trim support in the filesystem (which is needed to allow Dhara to garbage collect). See devflash.c and rawflash.c here: https://github.com/EtchedPixels/FUZIX/tree/master/Kernel/platform-esp8266

PR #804, opened 2020-12-25
Atari ST vt boot fixes

frno7:
A few minor updates were needed to start on the Atari ST using GCC 10.2.


EtchedPixels (commented):
VT_BOTTOM can be turned into a #define of a variable

frno7:
Indeed! I have some ideas on the virtual terminal implementation. I think we ought to make it crash-safe anyhow with cursor bounds checking etc.

PR #808, opened 2020-12-25
gtags: Add target, gitignore and clean

frno7:


EtchedPixels:
Thanks for cleaning all this up

PR #818, opened 2021-02-08
Add HDIO_TRIM for use with flash devices.

davidgiven:
This adds an HDIO_TRIM ioctl to block devices which can be used to pass information about unused blocks to a flash driver. This is needed to make slow flash like the ESP8266's work (via an FTL). Blocks are trimmed by the filesystem when freed, and by the swapper when swapping in a process, as processes are only ever swapped in once and this ought to speed up swapping out.

It'd be possible to add support to devsd for this, issuing CMD38 commands to the card, but erasing's a bit more complex on SD cards and usually not necessary, so I haven't looked into doing that yet.

### DIFF
--- a/Kernel/dev/blkdev.c
+++ b/Kernel/dev/blkdev.c
@@ -72,14 +72,30 @@ int blkdev_open(uint_fast8_t minor, uint16_t flags)
     return -1;
 }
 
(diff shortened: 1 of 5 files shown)


EtchedPixels:
Thanks - merged.

PR #819, opened 2021-02-10
Add support for the ESP8266.

davidgiven:
This gets Fuzix running on an unmodified ESP8266 with SD card, in single-tasking mode.

This is based on the LX106 processor which is an Xtensa LX6 variant. It's got ~96kB of RAM, split into code RAM and data RAM, and no MMU. Because it's split I/D it needs a custom syscall_exec.c file, basic on syscall_exec16.c. It supports NAND flash via an FTL but it's not fast enough to swap to.

There are some kernel changes:

- fixing various assumptions that 32-bitness means flat 32-bit memory (plus lots of warning fixes).
- importing the Dhara FTL library into Kernel/lib (MIT licensed).
- modifying devsd.c to defer raising CS until the start of the next command for writes --- for some reason my cards would hang when spinning while waiting for write completion; this is potentially faster, too.
- an (optional) extra udata field to track the size of the text segment, needed for swapping out code.

Other than that it's a simple port. There are still bugs, including a mysterious register corruption in multitasking mode, but I can tackle those later...


davidgiven:
Thanks!

EtchedPixels:
Seems to have broken a few things - I will take a look

-I../dev/ -I../dev/net/ -c ../dev/propio2.c

../dev/propio2.c:245: error 78: incompatible types

from type 'unsigned-char function ( ) fixed'

  to type '_Bool function ( ) fixed'

Makefile:41: recipe for target 'propio2.rel' failed


Looks like your changes broke a few things.

davidgiven:
Sorry about that --- I thought I'd tested by building the MSX platform but something obviously slipped through. I see you've fixed it; thanks.

PR #821, opened 2021-02-16
Add ARM Cortex-M0+ support.

davidgiven:
This adds userland support for the Cortex-M0+ CPU (as made famous by the Raspberry Pi Pico). It produces generic ELF PIE relocatable binaries, mostly because I couldn't make ucLinux's elf2flt build. I have a binary loader for this in a forthcoming PR.

There's nothing particularly exciting about this: system calls are made using the usual swi instruction, with the system call number in the argument.


EtchedPixels:
I'd just soldered up the pins on my Pico to have a go at a port 8)

PR #823, opened 2021-02-17
Fill unused sector records with 0xff

pnp2084:
JV3 unused sector records should be filled with 0xff.
This also fills in the last non-record byte to 0xff making the JV3 image writeable.

This is (somewhat) documented in: https://www.tim-mann.org/trs80/dskspec.html#:~:text=directory%20track%20instead.-,The%20JV3%20Format,headers%20and%20data%20can%20follow.

"If a sector header is not in use (free), its track and sector fields must contain 0xFF"


EtchedPixels:
Thanks - things I never knew 8)

PR #825, opened 2021-02-17
Simplify the ELF loader by quite a lot

davidgiven:
By changing the linker script to bake information about the size of the BSS and where the relocations are into the ELF file, I can remove all the section header stuff from the loader, which simplifies things considerably.

However, GNU ld then behaves very oddly if I try to strip certain sections (including rearranging or removing program headers... I think it could be a bug). This means I have to be rather less aggressive there. This doesn't affect the amount of data in memory, but it does make the files on disk a bit bigger.

### DIFF
--- a/Applications/rules.armm0
+++ b/Applications/rules.armm0
@@ -7,8 +7,8 @@ LINKER = arm-none-eabi-ld
 CFLAGS = -mcpu=Cortex-M0plus -ffunction-sections -fdata-sections -fno-strict-aliasing -fomit-frame-pointer -fno-builtin -Wall -g -Os -isystem $(ROOT)/Library/include
 LINKER_OPT = -L$(ROOT)/Library/libs -lc$(PLATFORM) -pie -static -no-dynamic-linker -z max-page-size=4
 LIBGCCDIR = $(dir $(shell $(CC) $(CFLAGS) -print-libgcc-file-name))
-LINKER_OPT += -L$(LIBGCCDIR) -lgcc -T $(ROOT)/Library/elfexe32.ld
-STRIP_OPT = -R .dynamic -R .dynsym -R .dynstr -R .shstrtab -R .hash -R .ARM.exidx -R .ARM.attributes
+LINKER_OPT += -L$(LIBGCCDIR) -lgcc -T $(ROOT)/Library/elfexe32.ld --no-export-dynamic -Bstatic -no-dynamic-linker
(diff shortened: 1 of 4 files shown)


EtchedPixels:
The likely 68K platforms all have at least 128K RAM (and the 128K one has a mandatory 64K kernel size due to the protection logic), the others can be in usually plentiful flash (PI, ESP8266, maybe one day pic32mx) - is it actually worth the deviation from ELF for them ?

It seems to be that either

1. You go with a true simple format and tight relocations and the ELF reader code you've written actually is the elf2fuzix tool, or 
2. You go ELF

otherwise it feels to me like falling between two stools - you have most of the complexity in the kernel and most of the incompatibility as well ?

davidgiven:
The program headers are explicitly there to provide a simplified mechanism which is easier for OS loaders to deal with --- the PT_LOOS to PT_HIOS range are dedicated to platform-specific extensions. So I don't think there's anything intrinsically evil about this. That said, the main rationale here was to avoid having to read in the section headers: as each entry is 40 bytes long, more than 12 section headers mean the section header table spans more than a single block, which the original loader didn't support. (And which can easily happen with unstripped binaries.)

I'm not sure that there's actually much more code here than in the FLT loader. OTOH, doing as much work as possible at compile time does make sense. But so does this simplicity of just being able to use ELF files... as you say, the ELF reader code here could be easily used at build time, so let me try and make a FLT converter using it.

frno7:
@EtchedPixels, I agree that keeping ELF compatibility is worthwhile, if it’s possible. A significant drawback would be that various ELF tools that one may want to use longer work as expected. Will it be possible with the ideas discussed here:

https://github.com/EtchedPixels/FUZIX/blob/3a068798ff535ba1d8de834ac449e419088931e8/Documentation/M68000#L89-L120

davidgiven:
Welp, I found a thing in the ELF spec which allows the loader to be dramatically simplified _again_ while also allowing the ELF files to be considerably more standard. So, no more custom program header types. (The secret is that PT_DYNAMIC points at a table where we can find the relocation table.)

The downside is that it requires the relocation data to be in an allocated section and read into memory alongside the program data and BSS. As each relocation entry is eight bytes, that adds up quickly. However, a lot of the problems I was having with the earlier loader is because ELF really, really wants the dynamic data to be mapped into memory, and GNU ld gets upset if you try to make it otherwise.

PR #828, opened 2021-02-20
Add prototype of a standalone tool for making bFLT files from static PIE elf files.

davidgiven:
This turned out to be pretty simple, but it currently won't work: the kernel loader needs a fair bit of work as it's currently hard-coded for the 68000, and the tool needs to be taught about 68000 relocations anyway. Also, it appears that bFLT is limited to 32-bit relocations only. I don't know how this works with platforms like the PowerPC where 32-bit constants are traditionally inlined but split into two 16-bit fields in two 32-bit instructions.

### DIFF
--- a/Standalone/elf2flt.c
+++ b/Standalone/elf2flt.c
@@ -1,0 +1,227 @@
+#include <stdio.h>
+#include <stdlib.h>
+#include <stdint.h>
+#include <string.h>
+#include <unistd.h>
+#include <fcntl.h>
+#include <sys/mman.h>
+#include <sys/types.h>
+#include <sys/stat.h>
+#include <arpa/inet.h>
+
+typedef uint32_t uaddr_t;
+#include "../Kernel/include/elf.h"
+#include "../Kernel/include/flat.h"
+
+#define perror_exit(msg) \
+	do { perror(msg); exit(1); } while (0)
+#define alignup(v,a) \
+	(uint32_t)((intptr_t)((v) + (a)-1) & ~((a)-1))
+
+#define ELFSTRUCT(offset) \
+	(void*)((uint8_t*)elffile + offset)
+
+static int stacksize = 4096;
+
+static void syntax_error(void)
+{
+	fprintf(stderr, "syntax: elf2flt [-s stacksize] -o outputfile inputfile\n");
+	exit(1);
+}
+
+static void write_error(void)
+{
+	perror_exit("cannot write output file");
+}
+
+void* load_file(const char* filename)
+{
+	int fd = open(filename, O_RDONLY);
+	if (fd == -1)
+		perror_exit("cannot open input file");
+	
+	struct stat st;
+	if (fstat(fd, &st) == -1)
+		perror_exit("cannot stat input file");
+
+	void* ptr = mmap(NULL, st.st_size, PROT_READ, MAP_PRIVATE, fd, 0);
+	close(fd);
+
+	if (ptr == MAP_FAILED)
+		perror_exit("cannot map input file");
+
+	return ptr;
+}
+
+int main(int argc, char* const* argv)
+{
+	const char* inputfilename = NULL;
+	const char* outputfilename = NULL;
+	for (;;)
+	{
+		int opt = getopt(argc, argv, "s:o:");
+		if (opt == -1)
+			break;
+
+		switch (opt)
+		{
+			case 'o':
+				outputfilename = optarg;
+				break;
+
+			case 's':
+				stacksize = strtoul(optarg, NULL, 0);
+				break;
+
+			default:
+				syntax_error();
+		}
+	}
+
+	if (!outputfilename
+			|| ((optind+1) != argc))
+		syntax_error();
+	inputfilename = argv[optind];
+
+	Elf32_Ehdr* elffile = load_file(inputfilename);
+	FILE* outfp = fopen(outputfilename, "wb");
+	if (!outfp)
+		perror_exit("cannot open output file");
+
+	/* Scan the section headers looking for stuff. */
+
+	Elf32_Shdr* shdr = ELFSTRUCT(elffile->e_shoff);
+	uint32_t bsshi = 0;
+	uint32_t bsslo = 0xffffffff;
+	uint32_t datahi = 0;
+	uint32_t datalo = 0xffffffff;
+	uint32_t texthi = 0;
+	uint32_t textlo = 0xffffffff;
+	uint32_t reloff = 0xffffffff;
+	uint32_t relcount = 0;
+	for (int i=0; i<elffile->e_shnum; i++)
+	{
+		Elf32_Shdr* sh = &shdr[i];
+		uint32_t seclo = sh->sh_addr;
+		uint32_t sechi = sh->sh_addr + sh->sh_size;
+		switch (sh->sh_type)
+		{
+			case SHT_PROGBITS: /* Initialised data */
+				if (sh->sh_flags & SHF_ALLOC)
+				{
+					if (sh->sh_flags & SHF_EXECINSTR)
+					{
+						/* Text segment. */
+
+						if (seclo < textlo)
+							textlo = seclo;
+						if (sechi > texthi)
+							texthi = sechi;
+					}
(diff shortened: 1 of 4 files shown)


Standalone/elf2flt.c
```diff
@@ -0,0 +1,227 @@
...
+
+	/* Assemble the bFLT header. */
+
+	struct binfmt_flat flatheader =
+	{
+		.magic = FLAT_FUZIX_MAGIC,
+		.rev = htonl(FLAT_VERSION),
+		.flags = htonl(FLAT_FLAG_RAM),
+		.entry = htonl(elffile->e_entry + sizeof(flatheader)),
+		.data_start = htonl(textlo + sizeof(flatheader)),
```
pawosm-arm, line 191: Is this really ok? Shouldn't it be `datalo` instead of `textlo`?

davidgiven: It's been a while, but FLAT combines data and text, with a single segment for both.

pawosm-arm: Something went wrong then. I've been trying to load these FLAT images with `syscall_exec32.c`. First it turned out for a little endian machine I had to add these:
```
        binflat.rev = ntohl(binflat.rev);
        binflat.entry = ntohl(binflat.entry);
        binflat.data_start = ntohl(binflat.data_start);
        binflat.data_end = ntohl(binflat.data_end);
        binflat.bss_end = ntohl(binflat.bss_end);
        binflat.stack_size = ntohl(binflat.stack_size);
        binflat.reloc_start = ntohl(binflat.reloc_start);
        binflat.reloc_count = ntohl(binflat.reloc_count);
        binflat.flags = ntohl(binflat.flags);
```
Then, the `valid_hdr()` function failed on this: `if (bf->entry >= bf->data_start)`.
As I replaced `textlo` with `datalo` here, it moved on, but it failed little bit further on this: `if ((bf->data_start | bf->data_end | bf->bss_end | bf->entry) & 1)`.
...turns out, `binflat->entry` is not even! And the value itself comes from this: `.entry = htonl(elffile->e_entry + sizeof(flatheader)),`. So I checked what `readelf` says about cross-compiled and linked binary:
```
  Version:                           0x1
  Entry point address:               0x11
  Start of program headers:          52 (bytes into file)
  Start of section headers:          103908 (bytes into file)
  Flags:                             0x5000200, Version5 EABI, soft-float ABI
  Size of this header:               52 (bytes)
  Size of program headers:           32 (bytes)
  Number of program headers:         4
  Size of section headers:           40 (bytes)
  Number of section headers:         22
  Section header string table index: 21
```
Well... entry point address is not even. I wonder what went wrong...

pawosm-arm: Taking another look, I'm seeing that this condition `if ((bf->data_start | bf->data_end | bf->bss_end | bf->entry) & 1)` is over-sensitive. `bf->entry` can be non-even when it's used in a call.

davidgiven: It's entirely possible that elf2flt is wrong --- it's never been tested. Also, the flt loader has builtin assumptions for the 68000 which need fixing for other platforms. Which one are you working on?

EtchedPixels: On 68K an odd entry address is a bug, probably the &1 should be &SOMETHING_ARCH_SET

pawosm-arm: @davidgiven it's armv7 (I've created the `Kernel/cpu-armm4` directory for it). I'm heavily reusing your work on `armm0` and the bringup code borrowed from nuttx. After a heavy struggle with SD card reader (something that even nuttx doesn't implement for this board), I'm finally able to load binaries from the on-card filesystem. Yet I didn't like the idea of the process memory slot used by `syscall_execelf32.c`, so I tried to force `syscall_exec32.c` to work. And there's a progress today! With those `htonl()` and `datalo` changes described above I'm seeing the right things at the right place in the memory.

@EtchedPixels I've rewritten this alignment check as such:
```
        if ((bf->data_start | bf->data_end | bf->bss_end
#ifndef __arm__
                | bf->entry
#endif
                ) & 1)
                return 0;
```
Finally, the `_doexec` routine has jumped into a binary loaded from the `/init` file! On the `bx r0` call instruction, an address in `r0` must be a target address plus 1 (hence the odd number; apparently, ELF file contains exactly the value to load into `r0`). I could finally see in gdb that PC landed in a code the same as `objdump -dS` shows for `init.debug` file.

Alhough `\init` has died eventually in the exception handler, for the first time it wasn't the HartFault!

pawosm-arm: The values `__bss_start` and `__bss_end` loaded to registers (in init's Wipe BSS code) are completely wrong, hence the exception. Seems I'll spend more time in `elf2flt`...

pawosm-arm:
A short update, there is a neat fix for the issue above in the `syscall_exec32.c` file:
```
@@ -82,7 +105,7 @@ static void relocate(struct binfmt_flat *bf, uaddr_t progbase, uint32_t size)
        uint32_t *rp = (uint32_t *)(progbase + bf->reloc_start);
        uint32_t n = bf->reloc_count;
        while (n--) {
-               uint32_t v = *rp++;
+               uint32_t v = ntohl(*rp++);
                if (v < size && !(v&1)) /* Revisit for non 68K */
                        /* FIXME: go via user access methods */
                        *((uint32_t *)(progbase + 0x40 + v)) += progbase + 0x40;
```
It makes `/init` continue much further, until it's hit with exception again, will look at it next time.

Standalone/elf2flt.c
pawosm-arm: A short update, there is a neat fix for the issue above in the `syscall_exec32.c` file:

```
@@ -82,7 +105,7 @@ static void relocate(struct binfmt_flat *bf, uaddr_t progbase, uint32_t size)
        uint32_t *rp = (uint32_t *)(progbase + bf->reloc_start);
        uint32_t n = bf->reloc_count;
        while (n--) {
-               uint32_t v = *rp++;
+               uint32_t v = ntohl(*rp++);
                if (v < size && !(v&1)) /* Revisit for non 68K */
                        /* FIXME: go via user access methods */
                        *((uint32_t *)(progbase + 0x40 + v)) += progbase + 0x40;
```

It makes `/init` continue much further, until it's hit with exception again, will look at it next time.

pawosm-arm: another mystery solved. In my `doinit()` implementation, I've been copying `u_sp` to `SP` while `syscall_exec32.s` sets only `u_isp`. This helped me to move on even further, past the first `putstr` at least. I got worried I didn't see anything on the serial terminal, but then I realized, there's no `/dev` in my filesystem! But then, how could it go pass the `do..while` loop trying to open `/dev/tty1`?! But that's another topic, I guess executables loading subject is now sorted. Lots of fun with it as long as I'm having fun :)

pawosm-arm: just for completeness, one more change was needed in `syscall_exec32.s`, `wargs()` function:

davidgiven: Does that mean we're now correctly loading elf2flt-generated binaries? Awesome.

Re ARM, I have a Raspberry Pi port somewhere I should brush off and make work again...

pawosm-arm: Yes we do:
```
FUZIX version 0.4pre1
Copyright (c) 1988-2002 by H.F.Bower, D.Braun, S.Nitschke, H.Peraza
Copyright (c) 1997-2001 by Arcady Schekochikhin, Adriano C. R. da Cunha
Copyright (c) 2013-2015 Will Sowerbutts <will@sowerbutts.com>
Copyright (c) 2014-2020 Alan Cox <alan@etchedpixels.co.uk>
Devboot
256kB total RAM, 256kB available to processes (15 processes max)
Enabling interrupts ... ok.
SD drive 0: hda: hda1
bootdev: hda1
Mounting root fs (root_dev=1, ro): OK
Starting /init
init version 0.9.0ac#1
init: no inittab
sh: Unknown error 2

panic: killed init
```
Still lots of work to do. And the filesystem was created all by hand.

Issue #192, opened 2015-05-24
Swap optimisation


EtchedPixels:
We need to move u_top into p->p_top. That also means we can make the swapper write out only the area required on a chmem app. However we then need some logic to always write out the udata/

davidgiven:
Assuming that the stack exists from p_sp to p_top isn't necessarily valid: it's legal for a process to change stack pointer, e.g. if it's doing threads or coroutines. If this happens, p_sp < u_break, and we've lost track of where the original stack actually begins. I suppose a workaround could be to save the entire user workspace in this situation.

PR #829, opened 2021-02-20
ELF loader fixes

davidgiven:
This fixes an embarrassing bug where the relocation data was never being discarded. It also moves the user stack (which with ELF has to have a hard-coded size) down between the bss and the heap. This allows the swapper to avoid saving everything from u_break to u_top, which improves performance substantially on the Pico. It also eliminates udata.u_top in favour of udata.u_ptab->p_top.

Fixes: #192


EtchedPixels:
I'd prefer to fix #192.

You don't in theory need to adjust the layout as nothing is in theory guaranteed between break and sp (give or take any redzone for the CPU) but I see no reason not to. In fact for some platforms it probably makes sense to do this (ELKS does it on 8086 for example and it would make some sense on Z180)

davidgiven:
Grr, wrote my message late at night... #192 is about the position of top, while this change is depending on the position of break. Should u_break be moved into ptab as well?

EtchedPixels:
This falls apart horribly because of all the asm uses of u_top and also because switching to udata.u->ptab->p_top turns fairly ok Z80 code generation into complete horse manure, and isn't ideal in some other cases. I guess p_top has to be a copy.

davidgiven:
It took long enough, but I finally got round to pulling out the merged u_top/p_top stuff and fixing all the merge conflicts.

EtchedPixels:
It seems to be adding a private rargs() wargs() back - any particular reason ?

davidgiven:
I did wonder about those! I thought I'd just somehow screwed up the merge and accidentally deleted them... fixed.

Issue #832, opened 2021-02-23
uget with ints


davidgiven:
So I saw the change to blkdev.c to remove the use of ugetl --- I didn't realise that wasn't available on all platforms. Unfortunately, I've discovered some other occurrences of the same bug, such as in `_pipe()`, which assumes the arguments are `int16_t`s. This was what was causing pipes not to work on both the ESP8266 and the Pico (and presumably on all 32-bit architectures).

However, just adding ugetl won't help because then it'll be the wrong size on 16-bit systems. There really needs to be a ugeti or equivalent, which is either 16 or 32 bits depending on the system.

I've managed to work around this for this specific case but I think there needs to be a more generic solution here.

EtchedPixels:
Agreed

davidgiven:
Sigh. You can tell I get to do this late at night, can't you? Turns out that ugeti/uputi already exist everywhere, so there is no work that needs to be done here other than to fix pipe(). Closing.

PR #833, opened 2021-02-23
Fix pipe() on 32-bit systems.

davidgiven:
This fixes the issue where pipe() assumes its parameter to be a `int16_t*`, causing pipes not to work on 32-bit systems.

See: #832

### DIFF
--- a/Kernel/syscall_fs.c
+++ b/Kernel/syscall_fs.c
@@ -291,21 +291,22 @@ arg_t _close(void)
 pipe (fildes)                    Function 40		?
 int fildes[];
 ********************************************/
-#define fildes (int16_t *)udata.u_argn
+#define fildes (int *)udata.u_argn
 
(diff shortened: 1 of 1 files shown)


EtchedPixels:
I think your plan for ugeti() makes the most sense here. It's easy enough on 8/16bit as it has no cost (it's ugetw) and on 32bit its ugetl() so also free.

davidgiven:
It turns out that ugeti/uputi already exist, so the fix here is trivial. I also had a quick scan for other occurrences of ugetw/uputw in contexts where they should be using the int versions, and didn't see any.

tormodvolden:
For the sake of us (or me) conveniently following the master (and e.g. future use of git blame) could it please be considered to rebase such pull requests (you can repush them) before they go into master? In any case, thanks a lot for all the development.

PR #835, opened 2021-02-25
Have video memory be linear

gp48k:
Bit 7 of $84 is the video page bit.  Unintuitively it specifies which page is second.  In normal operation it is set to 1 so that when video memory is mapped to $F800 you have page 0 at $F800 and page 1 at $FC00.

This change has been verified on a real machine.  It makes no difference on xtrs which is a bug in its emulation.

The boot block sets the high bit of $84 so no change is needed there.


EtchedPixels:
Thanks

PR #840, opened 2021-03-21
m68k: let globally overwrite the toolchain prefix using CROSS_COMPILE:

piso77:
m68k: let globally overwrite the toolchain prefix using CROSS_COMPILE:

$ make CROSS_COMPILE=m68k-buildroot-linux-uclibc- TARGET=tiny68k
    
If CROSS_COMPILE is not defined, it will default back to m68k-uclinux-


EtchedPixels:
Thanks

PR #852, opened 2021-07-10
[dk-tm4c129x] Initial commit for dk-tm4c129x support

pawosm-arm:
This PR adds initial support for the dk-tm4c129x board. It allows to build a filesystem image which can be stored on a SD card and the bootable kernel which can be loaded using JTAG, as shown on this video: https://www.youtube.com/watch?v=50dS0PzTths

The initial idea was inspired by platform-rpipico, some of the code was borrowed from other FOSS projects, namely, GCC and NuttX operating system. This massively simplified preparation of the board bringup code.


EtchedPixels:
NuttX is a problem as it's Apache licence and Apache 2.0 and GPLv2 don't mix. Most of the stuff seems to actually be 3BSD so that's less of a concern but the apache bits have to go. I also don't want to import half of nuttx or anyone elses SDK as ot gets unmanageable rather fast.

Otherwise it looks pretty sane on a first scan but reverts things in the base code for no reason (eg uint_fast8_t -> uint8_t in one spot I noticed)

pawosm-arm:
licensing problem.

EtchedPixels:
I've been through everything that's not in Kernel-platform* and merged most of it so that should help cut the tree back down to size.

EtchedPixels:
Yeah I had a look this evening. I've pulled the relevant bits out and redid it as it's not much it needed. Probably I broke some oddments so could do with a check

PR #859, opened 2021-08-11
Add in lots of Raspberry Pi Pico enhancements

davidgiven:
This contains changes to:

- move the stack to the bottom of the heap, not the top, allowing brk() to resize processes;
  - fix a very old bug in sh which assumes that the stack is above the heap...
- implement a much better swapper which divides the RAM into 4kB chunks, and does context switching by shuffling these around, only spilling to swap when needed; this allows the system to run comfortably with a NAND flash root file system and no swap, which means no SD card is needed
- the console is now exposed via a USB serial device as well as on the UART, which means you don't even need a UART adapter either

There's still no preemption but it's pretty responsive in cooperative mode.


EtchedPixels:
Thanks - there are a couple of things I think want tweaking but I'll do those after

PR #875, opened 2021-09-12
coco3: get it building again

beretta42:
networking off until further net dev.  this agets coco3 building again


EtchedPixels:
Thanks

PR #876, opened 2021-09-25
dk-tm4c129x: WIP: let's blink some lights on the network device

pawosm-arm:
This is still very far from being usable, but at least some
signs of life can be observed: once breakpoint is set on eth_isr()
in the debugger, one can 'ping' this device's hw address (e.g.
using macping.sh script) and the breakpoint on the IRQ handler
should be hit.

The `macping.sh` script I'm using is:

```
#!/bin/bash

network=192.168.1.1/24

if [ "$#" -ne 1 ]; then echo Usage example: $0 aa:bb:cc:dd:ee:ff; exit 2; fi;

nmap -sP $network >& /dev/null
ip=$(arp -n | grep $1 | awk ' { print $1 }')
ping $ip -n -q -c 2 -i 0.2 -w 1 >& /dev/null
if [ $? -eq 0 ]; then
    echo Device is online \($ip\)
else
    echo Device is offline
    exit 1
fi;
```
you need to change the `network` variable to the address space used in your home network.

### DIFF
--- a/Kernel/cpu-armm4/cpu.h
+++ b/Kernel/cpu-armm4/cpu.h
@@ -95,6 +95,8 @@ typedef union {            /* this structure is endian dependent */
 
 #define ntohs(x) ((uint16_t)(__builtin_bswap16((uint16_t)(x))))
 #define ntohl(x) ((uint32_t)(__builtin_bswap32((uint32_t)(x))))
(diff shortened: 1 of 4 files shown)


EtchedPixels:
Merged - thanks

PR #882, opened 2021-11-07
Fix assorted bitrot and make the CI builds mostly work again

davidgiven:
This fixes a few minor bugs which were causing the CI builds to fail. A few other notes:

- I've also added the ESP8266 to the build matrix.
- I have no idea why sbcv2 is still failing. The error makes no sense and I can't reproduce that locally. An sdcc bug, maybe?
- the z80 platforms take nearly _two hours_ to build, which is ridiculous.


EtchedPixels (commented):
(deleted)

EtchedPixels:
I've no idea what is going on with Z80. It does take a while to do a full build but nothing like 2 hours. I assume it's being dumped into spare cpu cycles at low priority somewhere which is fair enough. Z80 user space only needs building once though and for coverage testing the user space can be built with -O0 forced which is way faster because it avoids sdcc's incredibly expensive optimisation passes.

PR #883, opened 2021-11-08
One last piece of CI tidying

davidgiven:
This makes the esp8266 build go green. Still no idea about the sbcv2, though.


EtchedPixels:
No idea on the sbcv2 one - that's just weird

PR #884, opened 2021-11-09
Streamline CI and switch to a build matrix

piso77:
Stop cut&pasting pieces of code around, and switch to a build matrix - this should make the entire process easier to reuse and push forward.

And while here, give the file and action a more sensible name.


EtchedPixels:
Thanks

PR #885, opened 2021-11-09
w5x00: fix build failure caused by missing cast

piso77:
Fix sbcv2 build, with this one the entire CI pipeline is finally green.

Signed-off-by: Paolo Pisati <p.pisati@gmail.com>


EtchedPixels:
Makes sense - so it's a compiler version behaviour change

PR #888, opened 2021-11-14
Fix #887: Lack of bound check while loading elf32 segments

### LINKED ISSUE #887: Lack of bound check while loading elf32 segments
I noticed a lack of bound check while loading elf32 segments.

[https://github.com/EtchedPixels/FUZIX/blob/master/Kernel/syscall_execelf32.c#L208](https://github.com/EtchedPixels/FUZIX/blob/master/Kernel/syscall_execelf32.c#L208)

I currently run FUZIX on RP2040 platform which does not have any memory protection. But I supposed it is not an expected behavior as it would allow an arbitrary write with kernel privileges on platforms supporting MPU or MMU.

If you dont mind I will propose a pull request for this issue.

chatrasc:
This fix should prevent arbitrary write when loading elf32 segments. I previously reported it on #887.

### DIFF
--- a/Kernel/syscall_execelf32.c
+++ b/Kernel/syscall_execelf32.c
@@ -211,6 +211,13 @@ arg_t _execve(void)
 			udata.u_base = (uint8_t*) base;
 			udata.u_sysio = false;
 
+			if (base < PROGLOAD || (base + ssize) > (PROGLOAD + lomem)){
+				#ifdef DEBUG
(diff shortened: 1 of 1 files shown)


EtchedPixels:
think it also needs   base + ssize < base as an error check to avoid overflow cases ?

chatrasc:
You are right. You can decline this pull request and fix it your way.

PR #889, opened 2021-11-23
Add the framework of a ESP32 port.

davidgiven:
This does the basic work in getting the libc, applications and kernel up and running and to add it to the build matrix. There's not much kernel functionality yet but there's enough to bring up both cores, run fuzix_main on the application core, get as far as trying to mount the root filesystem.

The system call interface is based on an educated guess as to how it will eventually work in real life and may be subject to change. It uses the Fuzix flat binary format, just like on the ESP8266, as binaries can only be loaded at a fixed address, but ESP32 binaries are not compatible with the ESP8266 due to the different memory layout and CPU.


EtchedPixels:
Added

davidgiven:
Thanks!

PR #891, opened 2021-12-07
zx+3: make sure TERM is set to vt52 on ZX Spectrum +3

pawosm-arm:
It became annoying I had to edit /etc/inittab on each fresh +3 image.

### DIFF
--- a/Kernel/platform-zx+3/fuzix-platform-zx+3.pkg
+++ b/Kernel/platform-zx+3/fuzix-platform-zx+3.pkg
@@ -1,0 +1,7 @@
+package platform-zx+3
+
+disable-pkg platform-zx+3
+
+l /etc/inittab /etc/inittab.orig
+r /etc/inittab
+f 0644 /etc/inittab userspace/inittab
--- a/Kernel/platform-zx+3/userspace/inittab
+++ b/Kernel/platform-zx+3/userspace/inittab
@@ -1,0 +1,14 @@
+# Inittab
+id:3:initdefault:
+# Start a shell first
+# si::sysinit:/bin/sh
+# Run level s - a shell
+is:s:respawn:/bin/sh
+# Start up
+rc::bootwait:/etc/rc
+# Terminals
+01:3:respawn:getty /dev/tty1 9600 vt52
+02:3:off:getty /dev/tty2
+03:3:off:getty /dev/tty3
+# Shutdown
+ht:6:wait:/etc/rc.halt
--- a/Standalone/filesystem-src/build-zx+3-filesystem
+++ b/Standalone/filesystem-src/build-zx+3-filesystem
@@ -1,0 +1,16 @@
+#!/bin/sh
+
+endian=
+if [ "$1" = "-X" ]; then
+	shift
+	endian=-X
+fi
+if [ -z "$1" ] || [ -z "$2" ] || [ -z "$3" ]; then
+        echo "Syntax: $0 [ -X ] filename isize bsize"
+        echo ""
+        echo "For a 1.44MB floppy disk, use isize=64, bsize=2880"
+        echo "For a 32MB (or larger) hard disk, use isize=256, bsize=65535"
+        exit 1
+fi
+
+./build-filesystem-ng $endian -f $1 -g $2 $3 -p platform-zx+3


EtchedPixels:
Taught init to be smarter instead. That way the disk image is still shareable

pawosm-arm:
Thanks Alan, I'll give it a go when I'm back home next month (or I'll try it on some emulator if I find one suitable to use).
Merry Christmas!

pawosm-arm:
Still abroad, but I've managed to give it a run on Fuse emulator. It works nicely indeed and TERM is set to v52 as expected. But, I've encountered two regressions when I tried to build it:
1. `Applications/util/fleamacs.c`: the linker could not find definitions for `tputs()` and ` tgoto()`. A workarond was to comment out the use of both of them. Should a new issue be opened for it?
2. `Kernel/platform-zx+3/main.c` line 38: `strlen()` function return type mismatch (should be `int16_t` instead of `int`). Easy to fix.
And, in order to run it on Fuse, still the `CMD12` hack is required:
```
--- a/Kernel/dev/devsd.c
+++ b/Kernel/dev/devsd.c
@@ -204,7 +204,8 @@ int sd_send_command(uint_fast8_t cmd, uint32_t arg)
     /* Receive command response */
 /*    if (cmd == CMD12)  - ignore first reply byte anyway because it may
       be floating bus */
-    sd_spi_receive_byte();     /* Skip a stuff byte when stop reading */
+    if (cmd == CMD12)
+        sd_spi_receive_byte();     /* Skip a stuff byte when stop reading */
     n = 20;                             /* Wait for a valid response */
     do {
```

As a side note, I wanted to build the `zx+3` image with socket and bare network stuff included. It doesn't look like a huge piece of code, but it didn't fit anyway. Is there any advice on how this additional things should be laid in the kernel image?

EtchedPixels:
On Mon, 26 Dec 2022 04:13:40 -0800
pawosm-arm ***@***.***> wrote:

> Still abroad, but I've managed to give it a run on Fuse emulator. It works nicely indeed and TERM is set to v52 as expected. But, I've encountered two regressions when I tried to build it:
> 1. `Applications/util/fleamacs.c`: the linker could not find definitions for `tputs()` and ` tgoto()`. A workarond was to comment out the use of both of them. Should a new issue be opened for it?
> 2. `Kernel/platform-zx+3/main.c` line 38: `strlen()` function return type mismatch (should be `int16_t` instead of `int`). Easy to fix.
> And, in order to run it on Fuse, still the `CMD12` hack is required:
> ```
> --- a/Kernel/dev/devsd.c
> +++ b/Kernel/dev/devsd.c
> @@ -204,7 +204,8 @@ int sd_send_command(uint_fast8_t cmd, uint32_t arg)
>      /* Receive command response */
>  /*    if (cmd == CMD12)  - ignore first reply byte anyway because it may
>        be floating bus */
> -    sd_spi_receive_byte();     /* Skip a stuff byte when stop reading */
> +    if (cmd == CMD12)
> +        sd_spi_receive_byte();     /* Skip a stuff byte when stop reading */
>      n = 20;                             /* Wait for a valid response */
>      do {
> ```
> 
> As a side note, I wanted to build the `zx+3` image with socket and bare network stuff included. It doesn't look like a huge piece of code, but it didn't fit anyway. Is there any advice on how this additional things should be laid in the kernel image?
> 
> 

Ok so 1. Sill makefile error, fixed
2. Actually it should all be size_t, so I went back and fixed the
inconsistent Z80 ones

devsd - yep. Seems fuse is unmaintained, nobody ever responded to my
queries on this, or my patches to add support the Scorpion IDE. As it
breaks real hardware I'm not going to change it for a buggy emulator.

There are a few ways I can see to grab memory that I never got around to
doing

- The tty and disk buffers can be banked. I always hoped to do this so I
  could get user space much bigger by getting the kernel below C000 top,
  Can probably also stuff wiznet code in that space with care

- If you accept the C000 top of memory then the other choice (and maybe
  optional thing), would be to do what I'd wanted to do eventually with
  networking and make it loadable after boot. If it's loaded after init
  (eg by some kind of app) then you'd have about 12K free as the discard
  space would be gone.

Some of the work for that is done (it's why the net code communications
in only a tiny set of places) but there is a chunk more to do as well as
deal with the memory map juggling to get the chip mapped in.

pawosm-arm:
Hi Alan,

I'm sorry for disappearing for weeks between my messages, but that's the price of trying organize a life in a new place: I can't spend as much time on my hobbies as I wished to. It's a shame as I'm observing exciting changes being done to FUZIX and I can't participate (notably, I didn't try the GUI library yet; I wish I could port it to the Kentec K350QVG-V2-F display installed on board of my dk-tm4c129x kit, in theory I have enough documentation to do so, but it may take time before I'd be able to).

As for ZX+3 (it has earned permanent desk space again!), I can't say whether I can accept the C000 top of memory, as I don't know what cost or impact it would have (e.g. on the widely understood user experience). The readmes provide some ideas on what the memory map is or alternatively could be (so which is currently used?), but not giving much information what are the pros or cons of the given setup.

Making networking stuff loadable after boot seems obvious choice, yet what I struggled to fit in now, is the socket library which I guess is the more abstract thing, but maybe I'm wrong, maybe it should also be separated from the kernel core.

From other things, I saw you're doing train models and I have a few sets of tracks (still waiting to be unpacked, but it will reach its turn eventually). Thing is, it's the Z scale (all made by Maerklin), it doesn't seem very popular in the UK, but maybe you did something for that kind of gauge too?

pawosm-arm:
I just realized, what I meant was network syscalls (things listed in `network.rel` and `syscall_net.rel`), not the sockets, the sockets library is the userspace thing.

pawosm-arm:
I've find some time to read those mapping proposals again, and I got a feeling that there's some desperate attempt in there to organize direct video access from userspace. I guess it was inspired by one of the YT videos from 6 years ago showing how a film is played on 6809 board controlled by FUZIX. I doubt it would be ever possible on zx+3, so from my point of view, a more practical achievement would be some GSX compatibility layer (possibly with GEM VDI extensions to organize some GUI library on top of it, but that sounds like science fiction on this under-resourced platform), it doesn't care about direct access as everything goes through VDI's layers of abstraction and despite this, it wasn't that terribly slow on CP/M. Thing is, the whole idea of GSX seems to be so far away from the UNIX world, I can't see how it could be fitting well into FUZIX. Something like DirectFB would be a better choice and I guess the emerging UTK library is that thing.

Still coming back to the networking, I wonder how this "memory map juggling to get the chip mapped in" could be implemented? Now as the module loading facility is being implemented, I guess one feature of CP/M's loadable FID drivers would need to be considered: ability to hint where in memory the loaded module wishes to exist.

EtchedPixels:
There are some test bits I am playing with at the moment. Hence the changes in adding modlib.s / kmod-z80.s as a starting point.

Basically add an ioctl to ask the kernel for some memory, and a way to upload relocated code into it. For something like the +3 it's fairly simple to do the basicss for it

At boot discard gets blown away by buffers, and we end up with lots of buffers we don't really need. Post boot if someone asks for space for a module we sync the buffers to disk, and shrink the buffer pool down to leave room.

Just need to find the time to build the insmod tool and progress it beyond loading trivial proof of concept hello world modules.

pawosm-arm:
Hi Alan,

As my tm4c129x is currently broken, I'd look into zx+3 again. Are you planning to work on the insmod (as mentioned above) and the ability to fit a module into bank 5?

Also, I wonder if FUZIX gives any in-kernel ability to obtain a random seed of some kind. This is needed to init socket library in the Spectranet firmware.

Another thing, is FUZIX keeping some counterpart of CP/M's SVC_BANK_05 and SVC_BANK_68 system variables? They should contain the current contents of the hardware latches on channels #1FFD and #7FFD which is needed when fiddling with memory layouts and must be tracked by the kernel.

Thanks,
Paul

EtchedPixels:
I do need to get back to insmod but not sure when I will. The core stuff is there it's writing the bits to turn things into a usable format for relocations etc.

PR #892, opened 2021-12-08
net_native: make it build again (WIP)

pawosm-arm:
This is a preparation for a future work towards turning dk-tm4c129x ethernet device driver into a native driver.

### DIFF
--- a/Applications/netd/netd.c
+++ b/Applications/netd/netd.c
@@ -553,7 +553,7 @@ int dokernel( void )
 				/* FIXME: probably needs to do something here  */
 				/* fixme: set IP here? */
 				if (sm.s.s_type == SOCKTYPE_TCP)
-					m->port = sm.s.s_addr[SADDR_SRC].port;
+					m->port = sm.s.src_addr.sa.sin.sin_port;
 				ne.data = SS_BOUND;
 				ksend( NE_NEWSTATE );
 				break;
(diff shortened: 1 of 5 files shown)


pawosm-arm:
@EtchedPixels I'm getting slightly confused what the role of net_native should be: should it act as a regular network device driver (and hence implement e.g. netproto_read() instead of net_read()) or should it also replace network.c and therefore its current implementation of net_read() should be uncommented but also net_native should have its implementation of net_syscall() added? I would rather see net_native being a device driver similar to net_w5x00.c, but then some glue between network.c and net_native.c seems to be missing: most of the net_native.c functions are nowhere to be called!

Either way, my implementation of tm4c129x ethernet device driver waits for a major rewrite: it should not implement network device functions that network.c could call, it should expose its custom interface instead that the glue part of netd could open and utilize (something I started to implement in my other pull request).

EtchedPixels:
net_native still reflects the old API. I've just not had time to adjust it to the changes in the glue with everything that has been going on. In theory for bigger systems it would also be far better to either nail the uip code into the kernel flash image instead of having a usermode API, which was really originally designed to deal with "how to do i get tcp/ip in tiny 8bit system"

pawosm-arm:
> instead of having a usermode API, which was really originally designed to deal with "how to do i get tcp/ip in tiny 8bit system"

well, Fuzix is still targeting those systems so the userspace approach seems ok to me (although the file-backed communication looks weak, I'd rather move it to a ramdisk than have it going through SD card interface), e.g. ELKS has it that way and it targets slightly bigger 16-bit systems.

PR #893, opened 2021-12-09
dk-tm4c129x: attempt to turn dk-tm4c129x ethernet device driver into a native driver (WIP)

pawosm-arm:
Not much of a functional change yet, just preparing the ground.


pawosm-arm:
I suspect some if not most of the stuff in Kernel/platform-dk-tm4c129x/eth.c should be removed. We'll see, any hints are welcomed.

EtchedPixels:
This creates a problem if you have a simple system without select(). I had been thinking more along the lines of having the kernel<->netd interface also pass received packet/send packet, so you'd read() into the packet buffer, and if it was a notifier then process it, if not fire it into the stack, and if there was output fire it back up ?

pawosm-arm:
Yes, the kernel<->netd communication through a custom /dev/eth0 interface is something I'm thinking of and it will result in the removal of some of the code this commit currently adds. But meahwhile I've found how much netd needs to be updated to the current Fuzix reality, hence the work on this driver appears to be blocked by the required netd improvements.

PR #894, opened 2021-12-16
net_native: naïve rework that shuffles things around and makes a step towards further improvements

pawosm-arm:
This commit also renames netd-eth0 to netd-eth as the
previous name suggested there could be more than one
eth device. The /dev/eth0 was also renamed to /dev/eth,
for the same reason. Similarly to /dev/net, there can be
only one eth device present.


pawosm-arm:
I suppose the current device_* interface in netd is too simple to implement poll() or select() with timeout. Maybe it should be extended with some new functions with would result in extended list of the new /dev/eth ioctl requests?

PR #905, opened 2022-02-18
v68: fixed build and added diskimage creation

cnvogelg:
This PR allows to build a complete disk image (IDE image) for the Virtual68 platform.

I added the loader from Virtual68 to this tree so it can build all code of the image in one go. Additionally, IDE header and partition table for the 20M image described in the README were added.

Some minor fixes were applied to pass the build for v68: filesystem needs byte swap, assembler was misspelled.


EtchedPixels:
It's meant to be $(ASM) so that doesnt look wrong to me. Other stuff looks good at first glance

EtchedPixels:
Thanks merged

cnvogelg:
Indeed AS rules is not necessary at all. Thanks for merging and reverting the ASM change.

PR #909, opened 2022-03-08
dk-tm4c129x: fixes in the ethernet driver

pawosm-arm:
Three things were wrong:
- 'if' that was always true and was never needed blocked
  all of the outgoing traffic
- outgoing buffers were never freed
- /dev/eth was a block device while it should have been
  a character device

ICMP pings are now coming back from the device!


EtchedPixels:
Yay

pawosm-arm:
Hi @EtchedPixels,

Thanks for merging it so quickly. I've noticed that my commit message was slightly confusing and pushed slight correction, but you were faster and the commit with old message got merged :) Anyway, this is not a big deal.

PR #910, opened 2022-03-09
arm: fix huge instability in reading the process data from userspace

pawosm-arm:
With this patch it is finally possible to see the process
name on the arm-based devices.


EtchedPixels:
Eep thanks

PR #911, opened 2022-03-10
syscall_net: release closed socket to avoid running out of free sockets

pawosm-arm:
This solves a huge problem with userspace network related programs
that could be started only once, as any subsequent run was not
possible due to the lack of free sockets.

### DIFF
--- a/Applications/netd/netd.c
+++ b/Applications/netd/netd.c
@@ -115,7 +115,10 @@ void unhook_and_rel_map(int n)
 {
 	struct link *s = & map[n];
 	if (s->flags & LINK_UNBIND) {
-		ksend(NE_UNHOOK);
+		if (s->flags & LINK_DEAD) {
+			ne.data = SS_DEAD;
+			ksend(NE_UNHOOK);
+		}
 		s->flags &= ~LINK_UNBIND;
 	}
 	rel_map(n);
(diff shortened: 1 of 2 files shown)


pawosm-arm:
To reviewers: please share your opinion on possible data race on sockdata[i].socket between netproto_socket() and netproto_free()

pawosm-arm:
NB: I've added some basic protection of the disputed variable.

EtchedPixels:
This looks a bit odd to me - the userspace close of a socket doesn't mean the resources are free (consider stuff like TCP time wait). Unless I'm missing something this should be happening when the protocol module decides the socket is no longer relevant. Hence the comment.

pawosm-arm:
There must be something missing then. netproto_free() is only called from net_free() which in normal course of program execution is nowhere to be called. In effect, sockdata[i].socket (which links socket to socketdata) remains non-NULL even long after the process which assigned it is long gone. Very quickly netproto_socket() can't find free slots in sockdata[] array and no more sockets can be created.
Maybe syscall_net.c is not the right place to nullify this and netproto_free should be called from netproto_close()?

pawosm-arm:
Also, as I added call to `netproto_free(s)` to `case NE_UNHOOK:` in `netdev_write()`, it didn't help. Something more is missing.

pawosm-arm:
@EtchedPixels I've just pushed a new candidate, please take a look.

PR #913, opened 2022-03-20
network: several changes that finally made it possible to connect to remote hosts

pawosm-arm:
As connect() was never terminating, I looked at it and found several problems. Now connections are established and data can be transferred. Also emitting ICMP packets is fixed now so the remote hosts can be finally pinged.

A huge benefit from implementing in-memory ringbuf: netd-eth can now be started on read-only filesystems. The number of available sockets had to be halved though. Normally, 64k buffer would have to be allocated. Turned out, malloc() crashed for allocations larger than 4k, and a binary with a buffer larger than 32k in data segment cannot be loaded (it says it's too big to load).

Some issues are still remaining, for example:
1. httpd - listen() does not seem to work
2. ping `req=` shows suspicious values
3. telnet prints some trash after connecting to remote hosts; yet it works properly after that.

### DIFF
--- a/Kernel/syscall_net.c
+++ b/Kernel/syscall_net.c
@@ -136,7 +136,7 @@ arg_t _netcall(void)
 {
 	uint8_t flags = 0;
 	uint8_t cn;
-	arg_t *ap = udata.u_net.args;
+	arg_t *ap;
 	uint8_t op;
 	usize_t s;
 	int n;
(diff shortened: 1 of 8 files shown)


Kernel/syscall_net.c
```diff
@@ -136,7 +136,7 @@ arg_t _netcall(void)
 {
 	uint8_t flags = 0;
 	uint8_t cn;
-	arg_t *ap = udata.u_net.args;
```
pawosm-arm, line 139: this assignment was overwritten later. can you imagine the amount of debugging required to find it?!

EtchedPixels:
Thanks.. the memory buffer looks sensible although it probably has to be kernel side for most platforms because of swap.

PR #920, opened 2022-06-20
Fix some accumulated bitrot on the rpipico port.

davidgiven:
There was a pile of small issues which were preventing the Pico port from building. The biggest one (apart from the new source files) is the change in configuration in the USERMEM macros. I couldn't find a standard setting which worked --- `CONFIG_USERMEM_DIRECT` now only works on systems with no memory alignment requirements --- so I thought it simplest just to do a platform-specific `usermem.c`.

This now successfully boots an old userland I had lying around. It does not boot a current userland. I'll investigate that and try and find out what's wrong in another PR.

It also fixes some warnings.

### DIFF
--- a/Kernel/cpu-armm0/cpu.h
+++ b/Kernel/cpu-armm0/cpu.h
@@ -74,3 +74,6 @@ extern void copy_blocks(void *, void *, unsigned int);
 extern void swap_blocks(void *, void *, unsigned int);
 
 #define NORETURN __attribute__((__noreturn__))
+#undef __fastcall
+#define __fastcall /* nothing */
+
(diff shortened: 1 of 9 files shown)


EtchedPixels:
and that's broken esp8266 and esp32 it seems ?

EtchedPixels:
When I get a moment I'll untangle the mess I made for the usermem  direct stuff and the merge the relevant stuff on top.

davidgiven:
I, hm. I had completely forgotten about userspace alignment (it's been a while!). I _think_ that's probably a fair assumption --- we can't do anything about bogus pointers being passed to the kernel so we don't need to check for these either. There is a bit we can set which cases an actual exception rather than, I assume, bad data. Looks like `USERMEM_DIRECT` it is, then. Let me know when the fix is ready and I'll update the PR.

Re esp32 and esp8266: AFAICT these have always failed in the CI matrix. I don't see why. make is returning an error code but there don't seem to be any actual errors.

PR #929, opened 2022-07-03
Allow the ESP8266 port to run swapless.

davidgiven:
This backports the swapper from the Pico to allow multiple processes in memory at once with resizeable processes; it'll also use free pages in either code or data for storage, which makes more efficient use of RAM. I also managed to scrounge up an extra 16kB of RAM, which gives it just enough memory to boot to the shell without needing swap enabled. This means a NAND-flash-only configuration is possible (it also seems a lot faster, although I haven't actually measured it.) This requires an ABI change as udata now has to be at the bottom of user data.

Shared text pages would improve things a lot but I'm not sure Fuzix supports this.


EtchedPixels:
Nothing at the moment supports shared text pages but it's actually pretty trivial to support given a suitable architecture and memory map. You just keep a table of dev,inode for each executable running, and if the file is opened for write or truncated then clear the entry.

EtchedPixels:
Been busy - still pondering this. It's a big change in behaviour and a lot of complexity as well as clashing with the "so I added a PSRAM" patches I have somewhere

davidgiven:
There shouldn't be much change to the core code --- it's just adding p_texttop so that the swapper knows how much to swap in and out. The ESP8266 swapper code is a copy of the Pico code, except a bit more special because of the whole I/D thing it has going.

One day it'd be nice to refactor to common out some of the code but I don't think today is that day due to ESP weirdness.

davidgiven:
Also, re PSRAM: I don't know how your patch works but it should be straightforward to adjust my swapper so it treats the PSRAM as another bank of swap blocks --- there are already three. Blocks can be put anywhere; the only time the location matters is when a process gets swapped in. PSRAM's likely to be big so it'll want a large block size (4kB or so) to make the block location table reasonably small.

EtchedPixels:
Had a chance to take another look at this in detail.

Seems to make sense although it include/proc.h doesn't know about the new field on esp8266 . Needs a proper solution one day but for now I think adding the ifdef would do it

Second thing is that it would keep the old behaviour if it still automatically added the disk swap partition rather than just relying on swapon ?

Minor stuff though.

davidgiven:
Belatedly: thanks very much!

EtchedPixels:
Umm you did all the work so surely I should be saying thanks 8)

PR #936, opened 2022-08-12
Mods for RP2040 with alternative SD-Card pinning, and keeping kernel at build address for debugging

MarkMLl:
Please consider the attached as a tentative example. They are tested patches for an RP2040 running on a Cytron "Maker Pi Pico" board which use /dev/hdb2 on an externally-connected SD-Card as root, but they are not well integrated with the cmake build system more or less mandated by that platform.

### DIFF
--- a/Kernel/platform-rpipico/CMakeLists.txt
+++ b/Kernel/platform-rpipico/CMakeLists.txt
@@ -5,8 +5,21 @@ include(pico_sdk_import.cmake)
 
 project(fuzix C CXX ASM)
 set(CMAKE_C_STANDARD 11)
+
+# Do not enable this, leave the build type to be Debug, and assume that the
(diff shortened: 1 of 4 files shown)


erkinalp (approved):

MarkMLl:
I was discussing this with a journalist friend earlier who's writing on the new RP2040 Debug Probe (nee PicoProbe).

I have extremely limited skills with Git etc., but is there anything I can usefully do to help get these changes merged?

EtchedPixels:
There's a similar discussion been going on about the RC2040 so now a patch for that pushed which hopefully you can add to.

MarkMLl:
Thanks for the heads-up, I'll take a look but it certainly won't be for a few days.

PR #937, opened 2022-09-23
Cleaned up I/O-ports. 512KiB RAM, no ROM.

NollKollTroll:
These fixes and the latest addition to the pull request in RC2014 belong together to form the almost proper PZ1 emulation and real HW Fuzix environment.


EtchedPixels:
thanks

PR #940, opened 2022-10-04
appleiie: fix build

piso77:
BTW, i'm using vi and i noticed a lot of spacing issues, shall we enforce some spacing/style?


EtchedPixels:
If you want to clean it up I don't think anyone is touching the apple prototype work. It kind of died.

There is a CodingStyle document

piso77:
I tried the emulator linapple, but i couldn't get the fuzix.img to load at all - what have you used in the past to test it?
And how am i supposed to load it?

EtchedPixels:
It never got that far. Was just some initial scoping work.

PR #941, opened 2022-10-08
Pointer reassignment bug in floppy driver

xmotif:
When _devfd_init is called, GetPrm stores the drive table pointer in IY.
Eventually, Motor is called via Recal -> FdCmd -> Motor. All along it is assumed that IY is unchanged.
Within Motor, external function _plt_idle is called. For platform-rc2014, at least, the first thing that _plt_idle does is change IY to #_timer_source.
_devfd_init writes the value 1 to (IY) but coincidentally #_timer_source is often set to 1 so this bug does not cause crashes in most systems.

### DIFF
--- a/Kernel/dev/devfd_hw-banked.s
+++ b/Kernel/dev/devfd_hw-banked.s
@@ -478,7 +478,8 @@ FdcNotDn:
 ;
 ; Enter : None
 ; Return: None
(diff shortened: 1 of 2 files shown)


EtchedPixels:
That's a good find.. nasty bug

PR #945, opened 2022-12-05
Fix clock divider and print actual clock frequency on SC111

ra1fh:
CCR bit 8 needs to be set to switch the clock divider off according to the documentation. Verified by measuring PHI output frequency.

Z8S180-K has no clock doubler and CMR bit 8 is always 1. With that, try to detect missing clock doubler and print the clock frequency accordingly.

Before adding this fix, the clock speed was reduced to 9.2 MHz on my REV K Z180 and the serial port did run with 57600 bit/s instead of 115200. On a Z180 REV N with clock doubler, there may be a similar effect after adding this patch, it may speed up to 230400 bit/s. Just something to keep in mind when there's garbage on the serial port. I don't have a REV N to test with.


EtchedPixels:
Thanks. I'll have to find my SC111 and give it a test

EtchedPixels:
Thanks

PR #947, opened 2022-12-11
added power indication led

Darin755:
I found it hard to tell if the RPI Pico was on or not. This pull request turns on the power led as a indication that the board has power and has started to boot.


EtchedPixels:
Thanks

Issue #928, opened 2022-07-03
ESP8266: dc hangs


KedalionDaimon:
Assuming dc is the classical calculator, it does not seem to work at all and refuses to exit civilly or abort:

Welcome to FUZIX.
# dc -
2
3
*
p

q

help me the hell outta here

q
q
q

... And so forth, until I restart. Ctrl-C does not terminate it, unfortunately.

EtchedPixels:
See also #856

Issue #856, opened 2021-07-26
dc


EtchedPixels:
Debug dc command failures with Z80/SDCC

PR #949, opened 2022-12-14
Fix dc

palainp:
`hptr->beg` is freed just before being reallocated which can lead to a segfault (double free later). I have only tested this with esp8266 and it currently fixes https://github.com/EtchedPixels/FUZIX/issues/928 (maybe also https://github.com/EtchedPixels/FUZIX/issues/856 but I can't test due to lack of hardware).


EtchedPixels:
Thanks a lot.

PR #951, opened 2022-12-21
sc111: fix undefined reference to ds1302_init

ra1fh:
By default, CONFIG_RTC_DS1302 is not defined for SC111, resulting in undefined reference to ds1302_init.


EtchedPixels:
Thanks

PR #956, opened 2023-01-08
sc111: handle different SCM I/O base settings

ra1fh:
SCM < 1.3 did use I/O base 0x40. SCM 1.3 changed to 0xC0. Make sure to set the desired I/O base in both possible ICR register locations.

### DIFF
--- a/Kernel/platform-sc111/loader.s
+++ b/Kernel/platform-sc111/loader.s
@@ -43,6 +43,13 @@ start:
 	; because of the FFFE/FFFF thing.
 
 	ld sp, #0xFE00
+
+	; Adjust I/O base. SCM < 1.3 did use 0x40. Later versions
+	; changed to 0xC0.
+	ld a, #Z180_IO_BASE
+	out0 (0x7f), a
+	out0 (0xff), a
+
 	ld hl,#hello
 	call serstr
 


EtchedPixels:
That's going to break if the kernel expects 0x40 as it stil seems to and when the loader writes to the serial.

Probably simpler just to require SCM 1.3 or higher, or ROMWBW, move to 0xC0  ?

ra1fh:
The patch switches to the I/O base the kernel and the serial output functions expect. On my system this patch was required to make serial output work and boot Fuzix. I have SCM 1.3 (with 0xC0), and Fuzix running with default Z180_IO_BASE 0x40. I have tested with Z180_IO_BASE 0xC0 as well.

Requiring SCM 1.3 and changing Z180_IO_BASE to 0xC0 would certainly work as well. The release of SCM 1.3 happened fairly recently (Feb 2022) though. I don't know how many older systems with SCM < 1.3 are out there.

RomWBW has always been using I/O base 0xC0 on SC111. And the RomWBW loader for SC111 would do the I/O base switch if required before calling the kernel entry. So it does not have this problem.

ra1fh:
Replaced by PR #965

PR #958, opened 2023-01-23
sc126: fix wiznet polling

mennozon:
CONFIG_NET_W5X00 isn't defined and no data flow is possible without polling.

Fix this by checking if CONFIG_NET_WIZNET is defined instead like on other platforms.


EtchedPixels:
Thanks

PR #965, opened 2023-02-03
sc111: switch I/O base to 0xC0

ra1fh:
This primarily affects the SCM Fuzix loader, as it needs to know the current I/O base in order to work correctly. If there is a mismatch, Fuzix won't boot without any output on serial console.

With version 1.3.0 (2022-02-28), SCM switched from I/O base 0x40 to 0xC0. This change will make the SCM loader work on newer systems. Systems with SCM older than 1.3.0 will need this change reverted.

The RomWBW loader is not affected as RomWBW did always use I/O base 0xC0 on SC111 and the loader uses that knowledge to switch the I/O base if needed.

### DIFF
--- a/Kernel/platform-sc111/config.h
+++ b/Kernel/platform-sc111/config.h
@@ -35,7 +35,7 @@
 #define NMOUNTS	 4	  /* Number of mounts at a time */
 
 /* Hardware parameters : internal hardware at 0x40-0x7F */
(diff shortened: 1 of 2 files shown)


ra1fh:
This replaces PR #956

EtchedPixels:
Thanks.

PR #970, opened 2023-02-11
sc111: clear ASCI error flags after reading input characters

ra1fh:
This fixes a bug seen on Z8S180-K. When using GTKterm to paste a few characters to the serial port, the Fuzix system became unusable. Input data was still being processed and echo'ed back, but the wait pointer of the shell process got cleared and the tty driver would not deliver input data anymore. On further debugging, it turned out that the ASCI interrupt would trigger continuously with OVRN set and RDRF cleared when this happened.

Clearing errors by writing to EFR before exiting from the interrupt handler fixes the problem. According to the Z8S180 errata, clearing error flags might be a good idea on other CPU variants as well to prevent ASCI from stopping input processing after overrun.


EtchedPixels:
Thanks

PR #971, opened 2023-02-19
devtty: fix backspace echo not working properly with ASCI

ra1fh:
When backspace is read from the serial line, tty_inproc() will echo this back as "\b \b". This happens in interrupt context, so tty_putc_maywait() will not call tty_writeready(). On SC111, the space character got dropped because ASCI was still transmitting the first \b character. So only "\b\b" was transmitted on the serial port.

This was revealed by commit a56be113 ("sc111: now we have boot parameters don't hardware turbo mode"), which moved the DCNTL setup into a boot option, keeping DCNTL at the RomWBW/SCM default of 0 memory wait states on SC111. A minimum of 2 memory wait states is required on my SC111 to slow the CPU down to give ASCI enough time to transmit subsequent characters sent via tty_putc().

The check for TDRE in ASCI_STAT before writing to ASCI_TDR is present on several platforms already. Adding the check to all platforms that don't have it yet.

### DIFF
--- a/Kernel/platform-dyno/devtty.c
+++ b/Kernel/platform-dyno/devtty.c
@@ -132,9 +132,11 @@ void tty_putc(uint_fast8_t minor, uint_fast8_t c)
 {
     switch(minor){
         case 1:
+            while(!(ASCI_STAT0 & 2));
             ASCI_TDR0 = c;
             break;
         case 2:
+            while(!(ASCI_STAT1 & 2));
             ASCI_TDR1 = c;
             break;
     }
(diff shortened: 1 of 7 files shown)


EtchedPixels:
Unfortunately this is not the fix. There needs to be a bit of queuing or state to track pending erase. For backspace it's not hard but doing line delete either needs the old unix behaviour of a newline or something smarter.

Otherwise you end up hanging on the serial in the interrupt.

It's on the list of things to tackle properly at some point for platforms without serial queues.

ra1fh:
I should have called this workaround rather than fix :-) I understand that spinning in interrupt is not as designed, but until there is a serial queue this might be better than dropping characters.

PR #972, opened 2023-02-24
sc111: fix PRT timer interrupt flag reset

ra1fh:
To reset the timer interrupt, read TCR first and then TMDR0L. Otherwise, on Z8S180-K the timer interrupt triggers with double of the intended frequency.

### DIFF
--- a/Kernel/platform-sc111/main.c
+++ b/Kernel/platform-sc111/main.c
@@ -31,8 +31,8 @@ void z180_timer_interrupt(void)
     unsigned char a;
 
     /* we have to read both of these registers in order to reset the timer */
-    a = TIME_TMDR0L;
     a = TIME_TCR;
+    a = TIME_TMDR0L;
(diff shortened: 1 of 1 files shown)


ra1fh:
The official documentation is not very clear about the ordering. Some Z180 manuals contain a timing diagram that shows TCR first and TMDR second. I could only find one source about HD64180 that mentions that order is important:

https://oldcomputers.dyndns.org/public/pub/rechner/eaca/common/user-clubs/club-80/hardwarespezial(html)/hd64180-sonderinfo/hd64180-20-timer.htm

ra1fh:
This was found with ds1302 enabled btw. The time quickly ran away and started to jump centuries when the timer interrupt triggered too fast. With the fix in place, ds1302 works fine.

EtchedPixels:
Interesting little quirk. I'll go ripple that through the other Z180 ports

PR #979, opened 2023-03-26
dk-tm4c129x: fix linker error

pawosm-arm:
`LOBJ` are listed in `OBJS`, so there is no need to list `armrelocate.o` separately as it results in linker error:

```
Kernel/platform-dk-tm4c129x/../lib/armrelocate.c:10: multiple definition of `plt_relocate'; armrelocate.o:Kernel/platform-dk-tm4c129x/../lib/armrelocate.c:10: first defined here
```


EtchedPixels:
thanks

PR #980, opened 2023-03-26
NFC: these are not ARM's types

pawosm-arm:
Just random nit I've bumped on while looking at the most recent additions to FUZIX.


EtchedPixels:
The joys of cut and paste 8)

PR #983, opened 2023-04-03
Makefiles: help cleaning after itself

pawosm-arm:
A lot of executables was left behind when doing `make clean`.


EtchedPixels:
thanks

PR #992, opened 2023-05-07
platform-rcbus-z180: wiznet FDM mode for use on boards with no SPI chip select logic e.g. SC111

electrified:


EtchedPixels:
thanks

PR #999, opened 2023-05-30
MSX kbd improv

balyuf:
Note that the branch could be read as improvisation too ;-)

- It turned out that on MSX1 the tenth seconds where applied twice hence slowing down key repeat to molasses.  Fixed.
- Then took the repeat values from actual machines as a norm (which are not so linearly dependent on VDP frequency).
- The self-repeating ctrl-<sth> was due to only checking for SHIFT as modifier, changed that to SHIFT, CTRL, GRPH and CODE.
- Also, repetition stops not only on keys down change but also on modifier change.  Fixed.
- The big mystery of missing keys was due to me pressing the next key before letting go of the previous one.  Changed code to accept any number of keys down, yet only retain the last one pressed.  This is not the same as having multiple new keys down in one scan.  That will now give the highest scan numbered key as the so-called last pressed one.  This is consistent with actual MSX (though there they do input all pressed keys in repetition, which is a bit silly as they are of course in kbd matrix order and not in typed order, so no sense in trying to mimic this behavior).  But luckily for me, I am not that fast that my keystrokes end up in one scan! ;-)
- On JP keyboards we have an overlay where backslash is now the Yen sign.  Great, but until we have true i8n we should make sure that the function of that Yen key is the same as the backslash, as is the case in an actual MSX (e.g. path name separators are Yen signs).  So, in the key decode routine I now translate the Yen key to backslash, so that we can at least do something like echo -e "\0123" on the cli ...
- Finally there was an annoyance with recompilation for msx2 where the DSRCS where not properly handled with as consequence that some sources where recompiled each time even if they were not changed.  Fixed.
- Retro-actively fixed vdp1.s where there was left-over probably from copy/paste in clear_across.

### DIFF
--- a/Kernel/dev/vdp1.s
+++ b/Kernel/dev/vdp1.s
@@ -518,8 +518,7 @@ clear_across:
 	    ld a,(_int_disabled)
 	    push af
 	    di
-	    ld a, (_scrollu_w)
-	    ld b,a
+	    ld b, #0x40
 	    call videopos
 	    ld a, c
 	    ld bc, (_vdpport)
--- a/Kernel/platform-msx1/devices.c
+++ b/Kernel/platform-msx1/devices.c
@@ -115,8 +115,9 @@ void device_init(void)
     kprintf("%s.\n", vdpname);
 
     /* Default key repeat values in 10ths of seconds */
-    keyrepeat.first = 2 * ticks_per_dsecond;
-    keyrepeat.continual = 1 * ticks_per_dsecond;
+    /* These are the keyrepeat values observed for MSX1, irrespective of VDP frequency */
+    keyrepeat.first = 39;
+    keyrepeat.continual = 3;
 
     sunrise_probe();
 }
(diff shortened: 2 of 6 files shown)


balyuf:
Remark that we have JP overlay with YEN sign but on the screen there would be just a space as 0x84 is not attached to a Yen glyph, so we don't loose anything except weird behavior and we win our beloved backslash! ;-)

balyuf:
BTW, for the remaining MSX1 clear-across issue, you can easily verify that it has nothing to do with vi, by running this command, both on MSX1 and on MSX2 (or TurboR and aha now that will work with backslash ;-) ):

echo -e "test\013ab\033K"

This echo writes "test", performs CR, writes "ab", which leaves the line in a state of "abst" and then on MSX1 the last sequence does nothing and so "abst" is the end result, where on MSX2 the end result is "ab" ...

balyuf:
Ok, there was apparently a mistake in vdp1.s (and only there) with clear_across, probably a leftover from copy/paste or so. Fixed.

EtchedPixels:
Thanks a lot

PR #1001, opened 2023-06-06
msx2: make disk image bootable

balyuf:
MSX2 disk image is now bootable (hence no bootrom needed anymore) - that is if used with MegaFlash without romdisk, as I haven't found a key combination which disables booting from romdisk, except in the recovery mode (key up during boot) F3 which erases the romdisk (a bit drastic) ...

Would you consider publishing this MSX2 bootable diskimage on fuzix.org?

### DIFF
--- a/Kernel/platform-msx2/Makefile
+++ b/Kernel/platform-msx2/Makefile
@@ -40,22 +40,36 @@ $(DISCARD_DOBJS): %.rel: ../dev/%.c
 
 clean:
 	rm -f $(OBJS) *.lst *.asm *.sym *.rst *.lk  core *~
-	rm -f fuzix.ascii8.rom
+	rm -f fuzix.ascii8.rom fuzix.com boot.bin
 
 image:
 	dd if=../fuzix.bin of=fuzix.tmp bs=256 skip=1
 	dd if=../fuzix_disk.bin of=fuzix.com bs=256 skip=241
-	cp fuzix.com /tmp/com1
 	cat fuzix.tmp >> fuzix.com
-#	rm -f fuzix.tmp
 	dd if=../fuzix_boot.bin of=fuzix.ascii8.rom skip=1 bs=16384 conv=sync
 	dd if=../fuzix.bin of=fuzix.ascii8.rom seek=1 bs=16384 conv=sync
+	rm -f fuzix.tmp
+
+boot: boot.s
+	sdasz80 -o boot.s
+	sdldz80 -i boot.rel
+	# This makes us a binary from physical 0
+	makebin -p -s 49664 boot.ihx boot.tmp
+	# Chop off the leading 49152 bytes we don't want
+	dd if=boot.tmp of=boot.bin bs=512 count=1 skip=96 conv=sync
+	rm -f boot.ihx boot.tmp
 
 IMAGES = $(FUZIX_ROOT)/Images/$(TARGET)
 
-diskimage: image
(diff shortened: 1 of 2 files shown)


ricardojpinheiro:
It would be great if platform-msx2/Makefile had any way of using Sunrise IDE instead of MegaFlashROM SCC+ SD (what a long name, BTW). I could compile for MSX 2 and MSX 1, and I wanna use it into real hardware, but I don't have a MegaFlashROM SCC+ SD, but one Sunrise IDE compatible, a Tecnobytes IDE-Mapper interface (and 3 SD-Mappers, and 1 USB-Mapper...). 

I really don't know how to add this option into platform-msx2/Makefile (I'm a newbie at Makefiles). Any help would be appreciated. Thanks.

balyuf:
There was a comment here about using Sunrise IDE in MSX2 ...  I was thinking about that too, but it is more than just fiddling with the Makefile, as it also entails copying over some files from MSX1 and adapting them to the MSX2 platform structure, I am looking into it, but not sure yet if the adaptations will be trivial.  Maybe Alan can comment?  In the meantime we can perhaps merge in the bootable disk first and then I can create a new pull request for the Sunrise IDE (if feasible at all), as it may include changes to the diskimage creation (I seem to remember that an IDE images are not always interchangeable between the various disk technologies) ...

balyuf:
Thanks, looking into importing Sunrise IDE in MSX2, it seems the slot mapping and device tabulation stuff is not compatible, trying to understand the differences now ...

EtchedPixels:
Bankable kernel with the MSX2 mapper isn't really magical - it's probably not too hard beyond the extra complexity of the horrible MSX1 legacy MMIO mapping. The rc2014 kernel is banked and is fairly similar in mapping capabilities (any 16K page in any 16K address bank).

There are a few other things that will make space in future like moving to loadable modules for some stuff, switching to tinydisk/tinysd (not yet doable for multi-sd devices) and there are some things that probably would work now.

One is to turn on the rst peepholes and add a bit of code to set the vectors in the low memory of the kernel. See commit cfac1788a5e1416c0c4ccd655826d34e0fb5c4f7. That can be worth about 1K.

Another is to either load the font in the bootloader or just grab the font from the VDP and put it where it is needed (assuming the location is predictable or defined).

EtchedPixels:
Bankable kernel with the MSX2 mapper isn't really magical - it's probably not too hard beyond the extra complexity of the horrible MSX1 legacy MMIO mapping. The rc2014 kernel is banked and is fairly similar in mapping capabilities (any 16K page in any 16K address bank).

There are a few other things that will make space in future like moving to loadable modules for some stuff, switching to tinydisk/tinysd (not yet doable for multi-sd devices) and there are some things that probably would work now.

One is to turn on the rst peepholes and add a bit of code to set the vectors in the low memory of the kernel. See commit cfac1788a5e1416c0c4ccd655826d34e0fb5c4f7. That can be worth about 1K.

Another is to either load the font in the bootloader or just grab the font from the VDP and put it where it is needed (assuming the location is predictable or defined).

balyuf:
Thanks for the tips ... the rst peephole optimization indeed gives a code size just 95 bytes below 48k!  Struggling now to get the vectors right, some interrupt madness apparently ...  The good news is also that I have the Sunrise IDE working in MSX2, and it's blazingly fast in Turbo GT ...  So I should get things stable in time for 0.4 ...

PR #1011, opened 2023-07-06
Update Makefile.z80 for drivewire applications folder

ajcasado:
`make diskimage` for z80 target platforms complains about missing dwdate application, tested building for cromemco and zx+3.


EtchedPixels:
Not that we have drivewire on any of them but in theory its possible to do

PR #1036, opened 2023-11-13
Build and CI fixes

tormodvolden:
This makes the CI go green again.
And dragon-mooh build again (but not running since the post-v0.4 turmoil).


tormodvolden:
And now dragon-mooh runs again :)

PR #1037, opened 2023-11-13
CI: Add dragon-mooh with test-run in XRoar

tormodvolden:
<s>This needs a few improvements still, in order to not affect the normal builds.</s>
EDIT: Should be fine with normal builds now.


tormodvolden:
I don't see any CI builds on this PR. Did I screw up the yml, or does it need special approval since it is modifying that file?

EtchedPixels:
Thanks

PR #1042, opened 2024-01-17
Return error if tool check fails

tormodvolden:
And add tool checking file for esp32, fix the one for esp8266.


tormodvolden:
CI all green again!

EtchedPixels:
thanks

PR #1047, opened 2024-03-08
Make git fully aware of the case change

ChristianIvarsson:
Changing case of letters makes git rather confused if you don't perform it through git with the "git mv" command. Like "git mv Kernel\platform\platform-searle\tricks.S Kernel\platform\platform-searle\tricks.s"

You'll otherwise get some sort of Schrödinger situation where git sees both the old file as changed AND the new version on certain file systems.

I've manually checked that every "removed" file is the ghost of the old one before it was renamed but It'd be good if someone verifies just in case.


EtchedPixels:
This looks wrong.

A .S file is an assembly language file that is intended to go via the C preprocessor
A .s file is one that is not

This being standard Unix behaviour since forever.

There are a few cases where .s files lingered although they are now machine generated. I've slowly been tidying up.

You cannot build Fuzix on a case ignorant file system. It expects some vague degree of POSIX/Unix compliance.

ChristianIvarsson:
And cleaning up the names is what generated the problem to begin with.
There is a known problem with git where it'll generate a new entry instead of catching on that it has been renamed if you only change the case of the name.

If you do a clone of this version and compare files, you'll see nothing has actually been removed or renamed. - Only the old entry in the git database has been cleaned.

A simple way to make the problem manifest itself is if you make a clone with github desktop - You'll see this, pretty much unusable since it can't differentiate between the old file before it was renamed and the current one
gitweird

PR #1056, opened 2024-04-03
Truncate input values to their lengths

veremenko-y:
Truncate input values in `makeversion.c`. 
This prevents broken output of uname if values with exceeding length are suplied.

### DIFF
--- a/Kernel/tools/makeversion.c
+++ b/Kernel/tools/makeversion.c
@@ -2,6 +2,11 @@
 #include <stdlib.h>
 #include <string.h>
 
+/* Not sure why utsname.h needs <types.h>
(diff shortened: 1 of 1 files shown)


EtchedPixels:
How are we getting broken lengths in the first place is my question ?

veremenko-y:
I came across it when I tried adding timestamp to the kernel build, so I can track when last I compiled the kernel on the PI.
As of now I don't think there's any code path that hits it.

EtchedPixels:
Ok probably better to do something in kernel/start.c lilke

#ifdef KERNEL_PRINT_ID
      kputs(KERNEL_PRINT_ID);
#endif

PR #1057, opened 2024-04-03
Patched `ps` that reports used memory for Pi Pico

veremenko-y:
For pico every program is located at `PROGBASE`, so substracting it from `p_top` and rounding up to the ram block size (4k) will result in the correct value of memory allocated for process.


veremenko-y:
No worries. I understand why you're not sure. I think the only _proper_ way to keep track of it is to add it to `p_tab` which brings a lot consequences with it...

EtchedPixels:
Core stuff done. Just needs the target to set p->size

PR #1058, opened 2024-04-03
PI Pico: Save about 6K of RAM

veremenko-y:


EtchedPixels:
Added this one.. thanks

PR #1059, opened 2024-04-03
Pi Pico: Option to disable built in NAND flash

veremenko-y:


EtchedPixels:
-   timer_tick_cb(0);
+    hardware_alarm_force_irq(0);

Can you explain this part of the change. The rest makes sense to me

veremenko-y:
That line should have gone with #1060
Original line triggers callback in normal context, while the latter does trigger an actual IRQ. It is related to this code block
https://github.com/EtchedPixels/FUZIX/pull/1060/files#diff-32e7fc9e037bff07b2fb259d8074f48d5da28b0aa6a767181dd195d260aa1ab9R55

It might have been rushed decision, as I don't think anything was actually broken. On the other hand I felt like interrupt code should run in interrupt always, to avoid any unforeseen dependencies.

Not sure if that's  correct one, please advice. In the meantime I'll move it to the appropriate PR.

EtchedPixels:
Ok.. I will then tweak the define to fit the usual patterns

veremenko-y:
> If space is a premium might also be worth having a CONFIG_DHARA that all the dhara files are bracketed by and make devflash.c also compile to nothing without the flash enabled ?

When I tested it, I confirmed that linker successfully trimmed dhara as well.

PR #1060, opened 2024-04-03
Make USB uart separate from hardware UART

veremenko-y:
Also make hardware UART fully blocking for kput/kprintf

### DIFF
--- a/Kernel/platform/platform-rpipico/CMakeLists.txt
+++ b/Kernel/platform/platform-rpipico/CMakeLists.txt
@@ -27,6 +27,7 @@ add_executable(fuzix
 	main.c
 	misc.c
 	rawflash.c
+	rawuart.c
 	swapper.c
 	tricks.S
 	core1.c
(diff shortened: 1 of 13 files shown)


EtchedPixels:
-extern uint8_t progbase[USERMEM];
+extern char progbase[USERMEM];

This seems to be unrelated and strange ?

Rest looks ok but please try and use (void) not () for no arg functions. The PI compiler doesn't care but no some targets there is a difference between "no arguments" and "K&R style who knows arguments"

veremenko-y:
> Rest looks ok but please try and use (void) not () for no arg functions. The PI compiler doesn't care but no some targets there is a difference between "no arguments" and "K&R style who knows arguments"

Ack. I'll go though the patch to ensure it doesn't happen again. Do you want me to add this to the CodingStyle.md? (which I realized I haven't read, but it does missing the `foo(void)`).

> This seems to be unrelated and strange ?

This was part of fighting the compiler and conflicting types. I'll check this first and remove if unnecessary.

veremenko-y:
Both issues should be addressed now.

EtchedPixels:
Thanks - merged

Issue #1062, opened 2024-04-07
rpipico platform not working


Challmymind:
Newest pico build is not working. After flashing filesystem.uf2 then fuzix.uf2 there is no output on serial  (ttyACM0), sadly i cannot test if UART is working. I have done the same with the tarballs from [here](http://cowlark.com/2021-03-10-fuzix-pi-pico/index.html) with the success. Newest branch sets the LED on but as far as i can tell after it sits dormant. I will try to build version using previous commits before IOCTL changes (i dont know if it has broken things i just want to give it a try).

EtchedPixels:
I suspect @veremenko-y 's usb serial changes might be a more likely candidate

Does a682f7e38ae477c525b80670e7ea2d1127b3cd84 run as a starting point - that's just before all the recent changing around.

Challmymind:
You were right, a682f7e38ae477c525b80670e7ea2d1127b3cd84 builds correctly and everything works
I might try to revert changes to the USB and keep those from IOCTL, maybe it will also work (as it should) without any additional fixes.

Challmymind:
Update: shutdown is also invalid. 34ff65a1186027d45b63f88e443787799cff8695 is working until closed by shutdown command, after shutting it down it won't boot again.

veremenko-y:
Ugh... I'll look at it today. What's your set up? Just bare pico and USB serial?

veremenko-y:
> Update: shutdown is also invalid. [34ff65a](https://github.com/EtchedPixels/FUZIX/commit/34ff65a1186027d45b63f88e443787799cff8695) is working until closed by shutdown command, after shutting it down it won't boot again.

So I know what's happening with this one. Because you're using usb serial, it takes usually skips the starting messages. Even remount shutdown still corrupts file system for some reason. And after boot you get fsck prompt. If you type `n` or `y`, it'll continue booting. 

Left is serial, right is USB.
![Screenshot from 2024-04-07 17-07-14](https://github.com/EtchedPixels/FUZIX/assets/4127660/57e00a67-00d2-4e4a-a66b-6e886c1ca348)

I had the same issue before my changes as well. If you add a sleep call after `core1_init();` in main.c, you'll likely see full boot.

What I'm going to do is:
* add a check to wait until USB is initialized, and only after that continue booting
* add config to switch boot tty to USB one

[...]

PR #1063, opened 2024-04-15
rpipico: configurable ttys

veremenko-y:
**Fixes:** #1062

~~Because not everybody is able to connect UART as boot console for Pi Pico, I added the ability to reconfigure the order of tty's.~~

~~Amount of ifdefs has been growing a lot, so I added small `mktty` tool that parses config and generates `devttyconf.c` and `devttyconf.h` that maps devices appropriately. Format is very simple:~~

```
# Format: tty_dev tty_dev_number
usb 1
uart 1
usb 2
```

~~For now there's still a problem of USB ACM not connecting fast enough to get all boot messages printed. I could make it so it waits for any key before boot, but then unattended boot is broken. Should I add it as another option as well?~~

Version 2:

Stage 1:
On startup `devtty_init` will wait for USB host connection, and if present before timeout expires, it will configure USB consoles first, UART second. So boot messages will be printed using detected device.

Stage 2:
At root FS mount `plt_param` will get executed either from CMDLINE or from MBR.
Boot arg in the format `tty=[usb|uart]N,...` will be parsed and tty's will get reconfigured once again.

For example, given `tty=uart1,usb1,usb2` the following mapping will be configured
* `/dev/tty1` - `uart1`
* `/dev/tty2` - `usb1`
* `/dev/tty3` - `usb2`


EtchedPixels:
That does create yet more things for the user to get wrong, and makes it ever harder to have a build that someone can just try.

It's been easier on the other ports that do this as they can probe hardware or you end up typing stuff at the boot: prompt (so you could print to several and declare whoever you type at console).

Is it reasonable to wait a couple of seconds at boot and see if there is a USB connect and if not boot with serial as the console ?

veremenko-y:
> Is it reasonable to wait a couple of seconds at boot and see if there is a USB connect and if not boot with serial as the console ?

It is reasonable.
Unless you tell me to kill it, I'd like to keep the configuration of ttys. But I could instead set either default config that does `detect usb1 fallback uart1` or use `plt_param` with default `tty=detect` or something.
So in this case default build will pick up whatever active device is, but one can still configure whatever tty's (I'm also planning to add second hardware UART, that's why I hope you let mkconf stay)

EtchedPixels:
plt_param is probably the best option. You can also then set the parameters you want to indicate if there is a usb console and how many tty ports are present. See "setboot".

I really don't like the make config stuff, it's just guaranteeing a load more bug/queries from everyone trying to figure out yet another layer of indirection on configuring the device.

So my preference is strongly
- give USB a couple of seconds to connect, if it does it's the console
- if not assume serial

Let the user specify the number of ports on the command line (or with setboot).

So you'd then boot with USB, log in "setboot whatever" and it's done and consistent with the other ports.

veremenko-y:
> I really don't like the make config stuff

You're the boss :) So current suggestion is:
* By default and if `#define CMDLINE NULL`, wait DEV_USB_DETECT_TIMEOUT seconds, check if USB host is connected. If it is, put all USB devices first then UART devices into `ttymap`, otherwise the do the opposite and continue booting.
* Alternatively parse boot parameter like this `tty=uart1,usb1,usb2` which will fill in `ttymap`, and will not check for USB host
* `kputc` will use UART before the devtty is initialized, in case something panics.

----

I just took a look at the code once more, and `plt_param()` is executed after `device_init()` and right before `init` is executed.

So I would have to initialize devtty with defaults as I described above, then do it again inside `plt_param()` which I hoped to avoid. And I don't want to change start.c without discussing it first.

I will prototype it the way I described it, but it'll likely need to change based on your response.

veremenko-y:
Even more, `tty_init` is running way before anything else, and does not have a platform callback. Not including the fact that there's no access to `cmdline` yet at that point.

veremenko-y:
I pushed the prototype for you to take a look at.

It's not fully functional yet, I still have few issues with USB console.

veremenko-y:
Sorry for the delay, it does seem to work fine with default boot, but there are some issues with locking when changing the parameters in cmdline, and I've seen few Hard Faults running the system. I have a very busy week, but I hope to finish it the following week.

veremenko-y:
Who knew that using 5v SD card reader can lead to issues...

I think it's working now.

veremenko-y:
@EtchedPixels Sorry for a ping, doing it just in case if you missed the update. If not, I'm sorry, I wont' do it again :)

FYI something broke on pico platform when you updated `p_tab` struct, I get strange memory corruptions. I'm still investigating what it could be... I bisected it to 7fd6add4cc5a16beaf7bac9bf2826b2c4515f21b

EtchedPixels:
Weird.. I don't see anything else assuming the size and I imagine you did the obvious things like build from clean ?

veremenko-y:
But outside of the segfault issue, do you have any feedback for this PR specifically?

EtchedPixels:
Looks good to me other than pinning down the fault problem (which is probably unrelated anyway)

EtchedPixels:
Thanks

PR #1073, opened 2024-06-10
tinyide: fix IDE device detection on boot

ra1fh:
In my case (SC111 with SC145 and CF), the device signals DRDY before BSY is 0, which leads to a timeout and failed disk detection. So wait for BSY to be 0 in addition to DRDY before submitting the identify command.


EtchedPixels:
Thanks. Will get to it in a couple of days. That looks correct - in fact I don't think the state of DRDY is even defined when busy.

EtchedPixels:
Thanks

PR #1079, opened 2024-07-12
Gpio access on the Raspberry pi Pico

### LINKED ISSUE #946: Gpio access on the Raspberry pi Pico
Is there any plans to add gpio access in userland? I think it would be really useful to enable to access some of the low level hardware from a program running in userland.

I think the easiest way to do this is to create simple character devices that can can be toggled with a 1 or a 0.

napicella:
resolves #946

Small change to use the gpio interface with the pico.
The picoctl binary can be used to turn on/off the pins on the pico, e.g.:

```
Welcome to FUZIX.
# picoctl
Turn on/off GPIO pins on the pico
Usage: picoctl <PIN> <VALUE>
# picoctl 4 1
```
which turns pin 4 on.

It also the CONFIG for the [pico maker pi board](https://www.adafruit.com/product/5160)


EtchedPixels:
Thanks

veremenko-y:
@napicella I wonder why flash ioctl was removed?

EtchedPixels:
Oops missed that - split them apart and restored the old one so it's now two things

PR #1082, opened 2024-08-26
netd: fix /dev/eth colliding with later introduced /dev/rd0

pawosm-arm:
I've found it when looking into the `/dev` tree on dk-tm4c129x


pawosm-arm:
This is most likely insufficient. The kernel should also somehow know that the major/minor numbers have changed.

EtchedPixels:
I will take a look - at the moment they are both different device specific things

pawosm-arm:
> I will take a look - at the moment they are both different device specific things

I have a fix for it, just want to verify it this time before I push it

pawosm-arm:
Ok, I've tested with local network. The most recent FUZIX is still workable on this board.

PR #1089, opened 2024-09-25
netd: add ifconfig to Z80 build

electrified:
Make diskimage fails if ifconfig is missing (also we need it)
```
Processing package netd-base
ERROR ifconfig:13 file not found
....
ERROR Fix errors and re-run
```


EtchedPixels:
Thanks

PR #548, opened 2017-11-11
htget: quick fixes

beretta42:
un-decapitate content, add to 6809 + fs build script.

PR #1094, opened 2024-10-05
Fix htget saving duplicate content

electrified:
Bug introduced when #548 was merged (first commit wasn't dropped) - remove the duplicate fix


EtchedPixels:
Thanks

PR #1099, opened 2024-10-16
man: Added local search mode

veremenko-y:
For testing manpages in fuzix format


veremenko-y:
I'm not sure its necessary, just putting it here for BDFL to decide :)

PR #1100, opened 2024-10-16
rpipico: better uart init/deinit. fixed interrupt handling

veremenko-y:


veremenko-y:
I don't think I tested it well enough

Issue #1093, opened 2024-10-04
Wiznet issues


electrified:
Hey there,

I am facing some issues using the W5500 and rcbus-z180 implementation:

- The file saved by htget is mangled - it includes the start of the response repeated. (see example output later)
- dns lookups don't work (/etc/resolv.conf is set correctly)
```
dig google.com
timeout: Invalid argument
```

I initially thought this was going to be an issue with the wiznet FDM code we added last year (didn't do much testing at the time), so I invested in a buspirate and have been testing with that and get largely the same results, both in builds with VDM with CS connected and FDM modes and CS tied low.

One difference between EmulatorKit and the real hardware is on EmulatorKit htget keeps going indefinitely, wheres on the real hardware you get the same content repeated a couple of times before it stops.

Not sure if the dig and htget issues are related or not?

I understand you are busy with other things, but wondered if either of these issues rang any bells? Or if you have any debugging tips? (I think I need to add some logging to htget and work on understanding the SPI trace output)

Thanks!

htget output:
```
htget http://192.168.1.8/ t2.htm
.....................................................................................................................................................................................................................................................................................................................Interrupt
# cat t2.htm
<!DOCTYPE html>
<html>
<head>
<title>Welcome to nginx!</title>
<style>
html { color-scheme: light dark; }
body { width: 35em; margin: 0 auto;
font-family: Tahoma, Verdana, Arial, sans-serif; }
</style>
</head>
<body>
<h1>Welcome to nginx!</h1>
<p>If you see this page, the nginx web server is successfully installed and
working. Further configuration is required.</p>

<p>For online documentation and support please refer to
<a href="http://nginx.org/">nginx.org</a>.<br/>
Commercial support is available at
<a<!DOCTYPE html>
<html>
<head>
<title>Welcome to nginx!</title>
<style>
```
etc etc

EtchedPixels:
Nothing immediate.  Definitely used to work 8)

My first point of call would be to see if the syscalls are returning sensible stuff to htget or htget is buggy (or the new compiler is miscompiling something)

PR #1104, opened 2024-10-20
Some UDP fixes

electrified:
I'm on a quest to get dig (and UDP generally) working on Z180/Wiznet. As reported in #1093 it is not working.

This half solves it - UDP datagrams are now sent correctly. Fixed up dig and echoping a bit in the process of trying to diagnose what was wrong.

Continuing to work on figuring out why UDP responses to us aren't read correctly.

### DIFF
--- a/Applications/netd/dig.c
+++ b/Applications/netd/dig.c
@@ -3,6 +3,7 @@
 
 */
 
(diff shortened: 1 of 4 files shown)


electrified:
Thanks for merging.

PR #1106, opened 2024-10-23
CPC6128 and ZX+3 fixes

ajcasado:
CPC6128: _plt_reboot bug fix, albireo transfer speed optimization and renaming of things inherited from zx+3 port.
ZX+3: put plt_ide.h in platform folder as build fails without it.


EtchedPixels:
Thanks

Issue #1091, opened 2024-09-28
ESP8266 Panic - FATAL EXCEPTION 28 @ 40216769 with 00000200


ermiah:
@EtchedPixels' recent commits resolved all the problems I had in my other [issue](https://github.com/EtchedPixels/FUZIX/issues/1090) and now both the build and burn processes work as expected. 

By running `dd if=FUZIX/Images/esp8266/disk.img of=/dev/sda` I successfully created a bootable FUZIX SD card. Contrary to what the [documentation](https://github.com/EtchedPixels/FUZIX/tree/master/Kernel/platform/platform-esp8266) states, in my particular case, the microSD module's VCC needed to be connected to 5V (not 3.3V) for D1 mini to be able to find and boot the file system. 

After boot, the kernel panics with the following exception: 
```
Devboot
Allocated 27 disk buffers.
128KiB total RAM, 96KiB available to processes (15 processes max)
Enabling interrupts ... ok.
NAND flash, 1421kB: hda:
SD drive 0: hdb: hdb1 hdb2 (swap)
Mounting root fs (root_dev=17, ro): OK
Starting /init
 _---,_   FATAL EXCEPTION 28 @ 40216769 with 00000200:
|     ||   a0=40217BE5  sp=3FFF8688  a2=00000200  a3=00000200
| RIP ||   a4=00000000  a5=00040000  a6=3FFFF7EC  a7=3FFE8900
|     ||   a8=10000000  a9=00000200 a10=000000FF a11=3FFFA2CC
|.,~||/|. a12=3FFE8800 a13=00000000 a14=00000000 a15=3FFF926C
 
panic: fatal
```
![image](https://github.com/user-attachments/assets/7981af2b-0eb0-4c18-8e1d-a8bd242b3a1f)

Sidenote: The [documentation](https://github.com/EtchedPixels/FUZIX/tree/master/Kernel/platform/platform-esp8266)'s commands need to be updated too (`make -C Kernel/platform-esp8266 burn` to `make -C Kernel/platform/platform-esp8266 burn`).

EtchedPixels:
Won't be able to do anything on this with the current code until new year probably. Have to finish the complete redo of the room that's my office

Issue #1090, opened 2024-09-26
ESP8266 -  A fatal error occurred: No segment header found at offset 0000 in ELF file.


ermiah:
I am trying to run FUZIX on an ESP8266 device. I have found multiple problems with the build process so far. 
1.  Running `make TARGET=esp8266 kernel diskimage` gives this error: `drivewire.h:125 file not found`. 
Commenting out the following line in `Library/include/fuzix-include.pkg` resolves this:
`f 0644 /usr/include/sys/drivewire.h sys/drivewire.h`

2) Running `make -C Kernel/platform/platform-esp8266 burn` fails. `platform-esp8266` directory has been relocated to `Kernel/platform`, but the `update-flash.sh` file inside that directory has not been updated. 

To resolve this, I have replaced all `../../`s in `update-flash.sh` with `../../../`

3) `make -C Kernel/platform/platform-esp8266 fburn` results in the following error: `A fatal error occurred: No segment header found at offset 0000 in ELF file.` I don't have any idea how to resolve this one. 

A more complete log of this command: 
```
make: Entering directory '/home/**/FUZIX/Kernel/platform/platform-esp8266'
./update-flash.sh
Making 512 byte/block filesystem with normal byte order on device filesystem.img with fsize = 2590 and isize = 32.
Source file '../../../Applications/util/blkdiscard' not found
chmod: can't stat file 'blkdiscard': 2
# -s is the flash size in kilobytes; the kernel and boot loader takes the first 64kB.
../../../Standalone/mkftl -s 1984 -e 0x1000 -p 512 -g 10 filesystem.img -o filesystem.ftl
Number of physical erase blocks: 496
Maximum logical size: 1299kB (2598 sectors)
Used space: 1295kB (2590 sectors)
xtensa-lx106-elf-objcopy -I binary -O elf32-xtensa-le --change-section-vma .data=0x40220000 filesystem.ftl filesystem.elf
esptool elf2image filesystem.elf
esptool.py v4.8.0

A fatal error occurred: No segment header found at offset 0000 in ELF file.
make: *** [Makefile:86: fburn] Error 2
make: Leaving directory '/home/**/FUZIX/Kernel/platform/platform-esp8266'
```

EtchedPixels:
Source file '../../../Applications/util/blkdiscard' looks the relevant bit

EtchedPixels:
Seems right for make burn now but not fburn. Can't test on actual hardware as my stuff is all packed away

PR #1115, opened 2024-11-24
Permit fburn for esp8266

palainp:
Dear @ermiah, @EtchedPixels,
This PR want to continue the work #1090 and now make fburn should works fine (it needs an additional call to ld to generate the segment headers, maybe there is a clever way to add them).
As I don't use anymore the SD card reader for the root filesystem, and as now fburn works fine, I've booted correctly without the panic in #1091. @ermiah would you mind to try to boot on the flash drive? (it requires to recompile the kernel ~~with a new option~~ after an updating config.h)
Best.

### DIFF
--- a/Kernel/platform/platform-esp8266/Makefile
+++ b/Kernel/platform/platform-esp8266/Makefile
@@ -72,7 +72,7 @@ image.elf: kernel.ld addresses.ld $(KOBJS)
 	$(CROSS_CC) -T kernel.ld -T addresses.ld -flto -mlongcalls -nostdlib -o image.elf \
 		$(KOBJS)
 	
-filesystem.img::
+filesystem.img:
 	./update-flash.sh
 
 filesystem.ftl: filesystem.img
@@ -81,6 +81,8 @@ filesystem.ftl: filesystem.img
 
 filesystem.elf: filesystem.ftl
 	xtensa-lx106-elf-objcopy -I binary -O elf32-xtensa-le --change-section-vma .data=0x40220000 filesystem.ftl filesystem.elf
+	xtensa-lx106-elf-ld -Tdata 0x40220000 -o filesystem2.elf filesystem.elf
+	mv filesystem2.elf filesystem.elf
 
 fburn: filesystem.elf
 	esptool elf2image filesystem.elf
--- a/Kernel/platform/platform-esp8266/README.md
+++ b/Kernel/platform/platform-esp8266/README.md
@@ -19,8 +19,18 @@ for further file storage and swap.
 
 ## Building
 
+The kernel needs to know the boot device, therefore, depending on your setup,
+you will need to run one of the following. If your root device is the flash:
 ```
-make TARGET=esp866 kernel diskimage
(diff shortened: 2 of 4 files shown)


EtchedPixels:
Can you stick some info in on the blkdiscard build fail and lets deal with that separately not quietly lose it in an unrelated pull. Otherwise looks good to me at first glance.

palainp:
Of course, would you mind if I squash/rebase to keep only one commit?
The compile failure is about absent macros, and I'm unsure why this happens now with esp8266 :(

palainp:
I needed to update the default boot device selection because it was broken before, sorry for that. I also removed the blkdiscard part of the PR.

In the meantime I also found that blkdiscard.c compiles fine, but as it is absent from the Makefile.common file in Application/util, we can't upload it into our flash image. So I wonder if the best path is to remove it from update-flash.sh or if we can use Makefile.generic in Makefile.esp8266 for the applications, or if we can add it to Makefile.common. The later cases can have impact for other platforms but I'm not aware of the consequences.

ermiah:
@palainp Hi. 
I've tried `git pull` and these commands to rebuild the Kernel and Image, and a reprogram the device (I'm using an SD Card for the file system): 
```
$ sudo make TARGET=esp8266 kernel diskimage
$ sudo make -C Kernel/platform/platform-esp8266 burn
$ sudo dd if=Images/esp8266/disk.img of=/dev/sda
$ minicom -D /dev/ttyUSB0
```
I'm still getting a Kernel Panic. Am I missing a step? should I do "fburn" instead?

```
Devboot
Allocated 27 disk buffers.
128KiB total RAM, 96KiB available to processes (15 processes max)
Enabling interrupts ... ok.
NAND flash, 1421kB: hda:
SD drive 0: hdb: hdb1 hdb2 (swap)
Mounting root fs (root_dev=17, ro): OK
Starting /init
 _---,_   FATAL EXCEPTION 29 @ 402141FA with 00000200:
|     ||   a0=9FFFFEBC  sp=3FFF8608  a2=6000011C  a3=0001FF00
| RIP ||   a4=60000120  a5=00040000  a6=3FFFE5A4  a7=00000200
|     ||   a8=10000000  a9=00000200 a10=60000144 a11=3FFFA2CC
|.,~||/|. a12=00476206 a13=00000001 a14=3FFF893C a15=3FFF894E

panic: fatal
```

palainp:
Dear @ermiah , I don't use the SD card since the PR which allows me to put the file system on the flash part of my device. So yes, what I do is use `fburn` (flash the file system) in addition to `burn` (flash the kernel), but this requires a change to the default device used to boot, in `Kernel/platform/platform-esp8266/config.h` you need to uncomment https://github.com/EtchedPixels/FUZIX/blob/5ac8387bc7262981c3d6b518a83b38539a8204ca/Kernel/platform/platform-esp8266/config.h#L95 and then recompile the kernel (which will use the esp8266 flash instead of the SD card as the root fs).

Now, back to your issue, it seems that the kernel is unable to run `/init` properly, and as it works fine in the flash (at least for me), I suspect that the root problem is around the SD card.

PR #1116, opened 2024-11-27
Fix beep generation

ajcasado:
This do_beep() function can be easily ported to other platforms with the same PSG as the zx+3


EtchedPixels:
Thanks

ajcasado:
Thanks for the heads-up!!

PR #1118, opened 2024-11-27
Links created for dosdir, doswrite and dosdel utils in filesystem image. Fix type mismatch compilation error on cu.c

ajcasado:
About dos filesystem manipulation utilities, I tried dosdir with a FAT16 partition and got the error reserved != 1 error. @EtchedPixels do you know the right way to use this with a partition in the same disk as the root filesystem?


EtchedPixels:
Not really played with it except briefly on floppies.

veremenko-y:
Oops, sorry for cu.c error. I rely on github to tell me I broke other targets.

PR #1126, opened 2024-12-23
rc2014: adapt to new RomWBW memory layout

ra1fh:
RomWBW version 3.4 changed the memory layout, putting the HBIOS working copy into RAM bank 0. To prevent the Fuzix loader from overwriting HBIOS, store data for bank 0 temporarily in bank 4. Copy bank 4 to bank 0 immediately before starting the kernel, when no more RomWBW calls are needed.

This works with older RomWBW versions as well, which had HBIOS in nRAM-2 and lower banks were free to use.

Tested with RomWBW 3.2, 3.4, 3.5-dev.101 on Zed Pro.


EtchedPixels:
Thanks

PR #1129, opened 2024-12-28
Delete .export statement in CPU_Z180 part

ao-kenji:
Delete .export statement in CPU_Z180 part to prevent the following error
with new bintools and compiler.

```
    :
Symbol hw_irqvector
While processing: ../../cpu-z80u/lowlevel-z80u.o
exported but undefined
```

Also updates the symbol name in comment.


EtchedPixels:
Thanks

PR #1130, opened 2025-01-04
coco3: fix populating table of LBAs from CCPT table

six809:
Post- instead of pre- increment populating this table meant HDA1 referred to the 2nd partition, etc.

Brings the logic in coco3-specific td_plt_setup() in line with the usual approach in tinydisk_setup().


EtchedPixels:
Thanks

PR #1131, opened 2025-01-05
tinydisk: allow access to MAX_PARTth partition

six809:
Space in td_lba[][] is reserved for MAX_PART+1 partitions (index 0 being whole disk), so access to [MAX_PART] is allowed (and without it you can't access hda5).

Edit: May be that dropping MAX_PART to 4 is also the "right thing" here - I just noticed it wasn't allowing hda5 which was enumerated in the CCPT hook because MAX_PART is 5.


EtchedPixels:
Made this CONFIG_TD_MAX_PART so it can be overridden and corrected the off by one
Thanks

PR #1136, opened 2025-01-12
coco3: video optimisations

six809:
Any TFR between index registers can be optimised so long as current Z flag is not used (and even then if the destination is U or S).  e.g.

        TFR  X,U        ; 2 bytes, 6 cycles
        ; becomes:
        LEAU ,X         ; 2 bytes, 4 cycles

There are also a few places where a pointer is loaded, then replaced with the zero-offset contents pointed to.  6809 indirect addressing can be used here for savings vs extended addressing (the savings go away wrt direct addressing, but I'm pretty sure that's not used here).  e.g.

        LDX     _curtty         ; 3 bytes, 6 cycles
        LDX     ,X              ; 2 bytes, 5 cycles
        ; becomes:
        LDX     [_curtty]       ; 4 bytes, 10 cycles

RTS can be merged into any PULS right before by adding PC to the registers pulled.

Slight reorder of plot_char to save a branch.

clear_lines does not need to use the U register.  We can also test for zero characters to clear early as the result of the MUL.

In scroll_up and scroll_down, use stack ops instead of predec/postinc wrt U.  And swap source and dest registers in scroll_up to enable that.

### DIFF
--- a/Kernel/platform/platform-coco3/videoll.s
+++ b/Kernel/platform/platform-coco3/videoll.s
@@ -141,7 +141,7 @@ tfr_cmd:
 c@	sta	b@+1		; !!! self modify inner loop
 	stb	b@+3		; !!!  
 	bsr	vidptr		; U = screen addr
-	tfr	x,y		; Y = ptr to Height, width
+	leay	,x		; Y = ptr to Height, width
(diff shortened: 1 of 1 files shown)


EtchedPixels:
Thanks

PR #1137, opened 2025-01-12
Dragon MOOH video optimisations

six809:
Optimise TFR usage similar to coco3 patch.

Save 1 byte and 4 cycles computing address within font.

Use the B register within plot_fast to make some savings.  Add 128 to
destination screen address in Y so that subsequent offsest are 8-bit.

Same 128 byte offset in m6847_clear_across, as well as opportunistically
using a post-increment to avoid the need for LEAX 1,X.

Same 128 byte offset in m6847_cursor_on/off.

In m6847_scroll_up/down, use stack ops with U instead of predec/postinc.

Also, fix a possible (maybe not now, but in the future) font corruption;
use LSLB instead of ROLB in fast multiply by 8.


EtchedPixels:
Thanks

PR #1138, opened 2025-01-15
Add platform for Amstrad CPC with standard memory expansion used as ram disk for swap

ajcasado:


EtchedPixels:
Can you make the docs explain why there are two ports, which to use and what the differences are.
Also why the expansion is used as swap rather than keeping tasks directly in each mapping ?

ajcasado:
I've decided to create a separate port because, given the need to use the swap mechanism, I couldn't find an easy way to compile the code to support both disk and RAM usage. Do you have any suggestions for simplifying this and merging both ports into one?

The only difference between the two is the use of a RAM expansion as swap space.

As for using it as swap rather than directly keeping tasks in the mapping, that's because it doesn't seem feasible with the current mapping model. The expanded memory can only be mapped in 16k blocks starting at address 0x4000, which also removes the common area from the map (this is the mapping I use to implement the RAM disk for swap while executing code in the non-common kernel code area below 0x4000). Alternatively, it can be mapped in 64k blocks, replacing the entire memory space without any common area to rely on.  See this references:
https://www.cpcwiki.eu/index.php/Gate_Array#Register_MMR_.28RAM_memory_mapping.29
https://www.cpcwiki.eu/index.php/Standard_Memory_Expansions

I'm currently studying this latter 64k block switching mapping for another port, inspired by how memory is handled in the Cromemco port.

If you find these explanations satisfactory, I'll include them in the documentation.

Regards,

Antonio.

ajcasado:
I've thought it over and I think I can merge both ports into one by properly defining the swap mechanism in the config.h file and using the C preprocessor to select the parts of the code to compile for each case. What do you think?

ajcasado:
Well, the only thing I'm not sure about when it comes to merging both ports is how to conditionally compile the rd_cpcsme.s file based on whether the extended RAM is being used or not.

EtchedPixels:
Its a tiny asm file - any reason for not just linking it anyway at least for now ?

Rest makes sense. I guess CPC ultimately does want the thunked model

ajcasado:
Both ports merged, build with swap in RAM expansion of 512k or 1024k or with swap in disk partition selection in config.h. Changed Readme file to explain the use of extended RAM and how to build the desired swap option.

EtchedPixels:
Thanks

PR #1140, opened 2025-01-22
Applications/util/less.c: Added a cut-down less(1) pager.

DoctorWkt:
This is a start at a `less(1)` pager for FUZIX. It works but could do with a few improvements.


DoctorWkt:
I've added code to switch to the VT52 escape sequences if TERM=vt52. This works as far as I can tell (using `xterm` in VT52 mode).

Weirdly, I made a different modification which used `tgetent()` and `tgetstr()` to get the actual escape sequences from `/etc/termcap`. I backed it out as it crashed the kernel when running the `68knano` emulator.

EtchedPixels:
Do you still have the tgetent/tgetstr one anywhere so I can debug the kernel ?

PR #1141, opened 2025-01-22
Add the wish shell.

DoctorWkt:
This is an interactive shell with no scripting facilities. However you can use ctrl-D and tab to expand things like usernames, variables, complete file/dir names. It has command-line editing and aliases.


EtchedPixels:
Nice. I doubt we'll ever get tcl/tk on Fuzix somehow so the name clash should be fine.

PR #1146, opened 2025-01-31
Several improvements on floppy loader, snapshot loader, mass storage disk image creation and dosread utils.

ajcasado:


EtchedPixels:
Thanks

PR #1151, opened 2025-02-04
Fixed routine '_hd_xfer_out' in Kernel/platform/platform-trs80/trs80.s:

GmEsoft:
It hanged during boot on real hardware (Model 4P with FreHD) because a busy wait loop was missing after the OTIR instruction.

Tested with trs80gp and on real hardware Model 4p with FreHD.


EtchedPixels:
Thanks

PR #1153, opened 2025-02-04
Fixed function 'probe_gfx' in Kernel/platform/platform-trs80/discard.c, not working because of write-only registers.

GmEsoft:
Fixed function `probe_gfx()` in `Kernel/platform/platform-trs80/discard.c`:
- the registers `gfx_x`, `gfx_y` and `gfx_xor` are write-only; we can't get the value back;
- can differentiate between Tandy and Grafyx (original) by trying to write a byte outside the visible range: the Tandy (and the Grafyx clone by IanMav) can do it, but not the original uLabs Grafyx board;
- currently can't differentiate between the Tandy and the Grafyx clone: the Grafyx clone can't XOR the text and the hi-res graphics using the `gfx_xor` register (existing only on the Tandy board).

Tested with trs80gp and on real hardware Model 4p with FreHD and a Grafyx clone from IanMav.

### DIFF
--- a/Kernel/platform/platform-trs80/discard.c
+++ b/Kernel/platform/platform-trs80/discard.c
@@ -18,7 +18,7 @@ __sfr __at 0x8E gfx_xor;
 
    The control port differs as follows
    Tandy
-   0: set to tun graphics on, clear for alpha
+   0: set to turn graphics on, clear for alpha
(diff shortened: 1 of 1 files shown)


EtchedPixels:
Thanks

PR #1154, opened 2025-02-04
Added AZERTY keyboard support for TRS-80 Model 4/4p in Kernel/platform/platform-trs80/devtty.c

GmEsoft:
Added AZERTY keyboard support for TRS-80 Model 4/4p in `Kernel/platform/platform-trs80/devtty.c`:
- the AZERTY keyboard layout is enabled by uncommenting the line containing `#define AZERTY`;
- accented and special characters are replaced with other rather important symbols.

Tested with trs80gp and on real hardware Model 4p AZERTY with FreHD.

### DIFF
--- a/Kernel/platform/platform-trs80/devtty.c
+++ b/Kernel/platform/platform-trs80/devtty.c
@@ -9,6 +9,8 @@
 #include <devinput.h>
 #include <stdarg.h>
 
+//#define AZERTY // uncomment this for AZERTY keyboards
+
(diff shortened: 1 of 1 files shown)


EtchedPixels:
Looks sensible. Will take a further look at it when I have a bit more time.

EtchedPixels:
Added. I will move the AZERTY define into the config.h file for the target though so it's obvious where to find it.

PR #1155, opened 2025-02-04
Fixed 2 bugs in Kernel/platform/platform-trs80/trs80load.s

GmEsoft:
Fixed 2 bugs in `Kernel/platform/platform-trs80/trs80load.s`:
- Wait state generation not active during kernel loading from floppy;
- One missing byte at the end (255 bytes instead of 256).

Tested with trs80gp and on real hardware Model 4p AZERTY with FreHD.


EtchedPixels:
Thanks a lot. My model 4 has alas been in storage for a long time

PR #1156, opened 2025-02-06
Kernel/platform/platform-rosco-r2: A port of FUZIX to the Rosco r2 m68k SBC.

DoctorWkt:
I think this is ready for inclusion, but let me know if things need fixing. Thanks!


EtchedPixels:
Added - not sure why the uart config would screw up either unless it's using the wrong table for the clock driving the uart ?

PR #1158, opened 2025-02-06
Library/libs/sprintf.c: don't write nul if size == 0

six809:
In snprintf() and vsnprintf().

Allows use of *snprintf(NULL, 0, ...) without segfault.

### DIFF
--- a/Library/libs/sprintf.c
+++ b/Library/libs/sprintf.c
@@ -64,7 +64,8 @@ int snprintf(char *sp, size_t size, const char *fmt, ...)
 	string->bufpos = (unsigned char *) sp;
 	rv = _vfnprintf(string, size, fmt, ptr);
 	va_end(ptr);
-	*(string->bufpos) = 0;
+	if (size != 0)
+		*(string->bufpos) = 0;
 
 	string->bufpos = p;
 	return rv;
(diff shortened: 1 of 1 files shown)


EtchedPixels:
Think it's a shade more complicated to fix properly. Will have a look when I have a bit of time.

EtchedPixels:
Think it should if (size < rv) so that it does the right thing for overrunning strings too ?

six809:
EtchedPixels wrote:
> Think it should if (rv < size) so that it does the right thing for
> overrunning strings too ?

I think I tested and the normal case worked ok - as if bufpos was
left pointing at the last character, not the one after by the file io.
But absolutely make sure I'm not hallucinating memories like some sort
of crazed AI.

..ciaran

six809:
EtchedPixels wrote:
> will take a look in detail - also if bufpos is ending up stuck on the last
> char whether the length return is right in that case and stuff..

Ah no you're absolutely right - and I was hallucinating (or really was
fixated on the NULL case I wanted to work).  Yes, more required.

..ciaran

EtchedPixels:
Think we need to check size == 0 at the start of the snprintf routines, and pass size - 1 down for the other cases

commit 6b8aca75c989283a6c82c7ea74338473d0cac25d

PR #1160, opened 2025-02-07
Applications/util/less.c: This now uses tgetstr() properly.

DoctorWkt:


DoctorWkt:
I'm not sure why it's failing on the 6502 side. I'm also no sure why the other termcap programs like `tget` are not being built on the 6502 side, given that `Makefile.6502` includes `Makefile.common`.

EtchedPixels:
You were including curses.h which mean the termcap functions weren't prototyped - cc65 doesn't like that as it needs correct prototypes to get the calling right. Changed the #include and all seems good

PR #1161, opened 2025-02-09
Applications/s: Another vi-like editor.

DoctorWkt:
This is a vi-like editor, with yank and search functionality. It also has working o,O commands, which vile doesn't have. The Makefiles are set up for 68000 and Linux only to start with, but I'm sure it will work on several other platforms.


DoctorWkt:
P.S I just noticed that the escape sequences are built into screen.c. Can I do the Makefile trick as per levee and make s-ansi and s-vt52?

EtchedPixels:
Nak to this. It's pulled from code in Webb Miller's book. Prentice Hall appear to own the original rights and it's not at all clear in anything I have seen that this has changed or that it's actually legitimately public domain/CC0.

If you've got or found anything from Prentice Hall or Webb Miller to the contrary I'd love to know as we could then boot vile out in favour of s which would make me happy.

DoctorWkt:
Got it. I should have considered that earlier. I might look at Elvis, Stevie and heirloom vi next :-)

EtchedPixels:
Hacked the sizes up a little bit but it's too big for 8bit really.

PR #1163, opened 2025-03-08
init: askfirst

veremenko-y:
Adds `askfirst` action to the init process.

Rationale is to allow to define multiple ttys without forking the process and using memory. Especially on the systems where sharing code segment is not possible.

For `askfirst` items the following rules apply: 

`NN:level:askfirst:command`

Where `NN` - is a tty number init is going to way for action "Please press Enter".

Init will open `/dev/ttyNN` in pid 1 with `O_NONBLOCK` and set alarm for 1 second, periodically checking if there's a character and whether it's `\n`.
Once `\n` character is received, init closes the tty and spawns the process from the `command` field. When killed it returns to polling tty in pid 1.


EtchedPixels:
That will be hideous on systems that swap init out (quite a lot of them).

The way System 5 sort of solved this was on systems where select (or rather poll() in their case) was added they had a single daemon that did a poll for all the pending tty ports and then ran the getty.
For 32bit machines finishing up select() support for ttys and doing this either with or in init sounds like a better option.

veremenko-y:
Good points. I haven't really gave a thought about init not being swapped out. I wanted to check if this idea works, and it did. But I missed that.

Please close if you don't see any value in it. I might play with tty select.

veremenko-y:
Closing due to the good feedback about swapping.

PR #1170, opened 2025-04-05
Library/libs/crt0_6809_rel.s: fix stashing argv[] as __argv

six809:
Showed up running "calendar".


six809:
The same code is in Library/libs/crt0_6809.s; it probably needs attention there too.

PR #1174, opened 2025-04-21
Kernel/mm/bank16k.c: add new pages below existing common

six809:
Otherwise it loses pages.

Only really visible if you start playing with chmem.


EtchedPixels:
Fun - introduced ironically while rewriting the code to stop gcc miscompiling it. Sooner we can go gcc free the better

PR #1176, opened 2025-04-25
Fix RC2014 CH375 map/unmap user space

hechtus:
The user space was not mapped/unmapped correctly when ch375_rblock() or ch375_wblock() where called.

### DIFF
--- a/Kernel/platform/platform-rc2014/rc2014.s
+++ b/Kernel/platform/platform-rc2014/rc2014.s
@@ -1194,16 +1194,6 @@ snooze:
 	.globl _td_raw
 	.globl _td_page
 
-map:
-	ld	bc,#0x40BE		; 64 bytes port BE
-	ld	a,(_td_raw)
-	or	a
-	ret	z
-	dec	a
-	jp	z,map_proc_always
-	ld	a,(_td_page)
-	jp	map_for_swap
-
 _ch375_rblock:
 	pop	bc
 	pop	de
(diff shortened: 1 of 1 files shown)


hechtus:
For some reason I don't understand this seems to be still buggy. When doing `ps` the kernel crashes. The following implementation works for some reason:
```
_ch375_rblock:
	pop	bc
	pop	de
	pop	hl
	push	hl
	push	de
	push	bc
	ld	a, (_td_raw)
	push	af
	cp	#2
	jr	nz, _not_swapin
	ld	a, (_td_page)
	call	map_for_swap
	jr	_doread
_not_swapin:
	or	a
	call	nz, map_proc_always
_doread:
	ld	bc, #0x403E
	inir
	pop	af
	or	a
	jp	nz, map_kernel
	ret

_ch375_wblock:
	pop	bc
	pop	de
	pop	hl
	push	hl
	push	de
	push	bc
	ld	a, (_td_raw)
	push	af
	cp	#2
	jr	nz, _not_swapout
	ld	a, (_td_page)
	call	map_for_swap
	jr	_dowrite
_not_swapout:
	or	a
	call	nz, map_proc_always
_dowrite:
	ld	bc, #0x403E
	otir
	pop	af
	or	a
	jp	nz, map_kernel
	ret
```

EtchedPixels:
td_raw isn't in common space so the second load of it when we are in user mappings will be testing some random memory location I think ?

hechtus:
OK. Makes sense. This explains why my last commit is working. I'm happy with it.

EtchedPixels:
Thanks

PR #1177, opened 2025-04-28
Changed td_transfer() parameters from uint8_t uint_fast8_t

hechtus:
On the Z1013 platform which is compiled with fcc this fixes a signed/unsigned issue of the statement

    uint8_t dev = minor >> 4

fcc seems to get confused with 8/16 bit or signed unsigned which breaks the statement above. It's possibly a fcc issue.

### DIFF
--- a/Kernel/dev/tinydisk.c
+++ b/Kernel/dev/tinydisk.c
@@ -20,7 +20,7 @@ uint8_t td_unit[CONFIG_TD_NUM];
 td_xfer td_op[CONFIG_TD_NUM];
 td_ioc td_iop[CONFIG_TD_NUM];
 
-static int td_transfer(uint8_t minor, bool is_read, uint8_t rawflag)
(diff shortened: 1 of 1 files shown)


EtchedPixels:
An unsigned char should get promoted to int then shifted.  The compiler wasn't clearing the upper half of the value before the shift in some cases so shifted junk into the top of the resulting byte. I've pushed a compiler fix.

hechtus:
Thanks! Your commit seems to fix this particular issue. But there is still something wrong I need to investigate further.

hechtus:
Strangely, your fix seems to have side effects. Now I'm getting a "no /init" panic. That was not an issue before.

EtchedPixels:
Not surprised - it means the compiler is taking some quite different paths and helpers so it's probably showing up bugs elsewhere that I will need to trace when I get some time.

PR #1180, opened 2025-05-30
68hc11: repair broken build

veremenko-y:
Not sure how it buit before, but `sci.h` was missing. Also `fuzix68hc11` folder is required, otherwise `Library/tools/syscall_68hc11.c` breaks.


EtchedPixels:
Stuff not commited my end - fixed I hope

veremenko-y:
Confirmed fixed

PR #1182, opened 2025-07-04
Several improvements to the platform cpcsme, mainly in support for CH375/6-based mass storage devices (Albireo and Usifac).

ajcasado:
The driver for CH375/6 now supports multiple devices simultaneously, which can be managed via serial or parallel interface. This code can be reused for other platforms, although it's important to note that the Amstrad CPC uses a 16-bit I/O address space.

Specifically on this platform, it provides simultaneous support for three USB drives: two Albireo devices and one Usifac.

The CH375/6 initialization code has been moved to the discard area.

The I/O transfer code for Tiny Disk-type devices has been unified to save space in the common memory area.


EtchedPixels:
Thanks

PR #1185, opened 2025-08-23
Get CI running again

tormodvolden:
Move CI to Ubuntu 24.04 images so we should be good for some time.

Unfortunately while we had no CI working, the esp32 build started to fail  and the dragon-mooh run test doesn't boot up properly any longer, but the rest seems fine.

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -16,7 +16,7 @@ jobs:
       fail-fast: false # keep on building even if one target fails
       matrix:
         target: [appleiie, dragon-mooh, dragon-nx32, esp32, esp8266, multicomp09, rbc-mark4, rcbus-6502, rcbus-68008, rpipico, sc108, tiny68k, armm0-libc]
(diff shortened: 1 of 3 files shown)


tormodvolden:
So now it is only esp32 failing:
```
esptool.py v4.7.0

A fatal error occurred: No segment header found at offset 0000 in ELF file.
```

A little hard to follow that parallel build log, but I think this is what leads up to it:
```
xtensa-esp32-elf-nm image.elf >../../fuzix.map
xtensa-esp32-elf-objcopy image.elf -O binary image.bin
xtensa-esp32-elf-objcopy -I binary image.bin -O elf32-xtensa-le --change-addresses 0x40080000 image-absolute.elf
```

```
esptool --chip esp32 elf2image -o image image-absolute.elf
```
EDIT: esptool is now at 4.7.0 versus previous 2.8, this can play a role.   <---- This. See #1186
EDIT: Looks similar to https://github.com/EtchedPixels/FUZIX/issues/1090

